PR #30, opened 2013-03-25
Have created a Windows Installer Package of asn1c 0.9.24

StefanGu65:
Have created a Windows Installer Package of asn1c 0.9.24 (except unber/enber, that will not compile under that version using CygWin).

One .exe file contains the installer.

This may help other people and cut down on support requests...
I am willing to create these packages on a continuous base.

/Stefan


markusschaber:
@StefanGu65 Any chance you will update to the current 0.9.27 version?

Alternatively, you could share a more detailed readme how to build it on windows (e. G. makefile modifications, which cygwin packages you need etc...), and make a pull request for that documentation. :-)

Thanks in advance!

markusschaber:
@vlm: would you accept pull requests for a "native" windows build? Either using GCC, MinGW or VS?

StefanGu65:
Sure,

 

I can update the windows version, it’s just that I have not had any reason to as the one I have worked well for me.

 

I’ll be on the road for about 2 weeks, if you can hold on a bit, I should be able to update it after this.

 

The resulting app is self-contained and works natively on windows. Due to the placement of the skeletons in windows, you may have to use the –s parameter as described in my readme.

 

S

markusschaber:
I tried to build it with cygwin myself. After installing some missing packages, it builds fine, but "make check" fails with:
`GeneralizedTime.c: At top level:
GeneralizedTime.c:132:15: error: static declaration of 'timegm' follows non-static declaration
 static time_t timegm(struct tm *tm) {
               ^
In file included from /usr/include/time.h:173:0,
                 from GeneralizedTime.c:12:
/usr/include/cygwin/time.h:21:16: note: previous declaration of 'timegm' was here
 time_t __cdecl timegm (struct tm *);
`

I also noted that GeneralizedTime.c does not work with Non-GCC compilers on Windows due to the `#warning` pragmas.

StefanGu65:
Unfortunately my build machine had a failed HDD and I don’t seem to be able to find the correct version of autotools to get the .27 version work.

 

What versions of autoreconf and autotools did you use? It’s been a while and I normally use Cmake…

 

/Stefan

PR #52, opened 2014-12-11
Disable extra portability errors in automake

dwd:
Automake 1.12 complains about a missing AM_PROG_AR for non-POSIX archivers.

This is probably a bug in libtool, or automake, or something else. Whichever,
it means that autoreconf -i fails, which is plain irritating.


anatolyborodin:
See also #62 and #63

vlm:
Applied manually.

Issue #62, opened 2015-07-15
Compilation issue -- AM_PROG_AR missing


fclaerho:
Hello!

I get those errors while compiling the project (ubuntu 14.04, x64); automake is nice enough to suggest the fix :-)

	(...)
	automake: warnings are treated as errors
	/usr/share/automake-1.14/am/ltlibrary.am: warning: 'libasn1compiler.la': linking libtool libraries using a non-POSIX
	/usr/share/automake-1.14/am/ltlibrary.am: archiver requires 'AM_PROG_AR' in 'configure.ac'
	libasn1compiler/Makefile.am:8:   while processing Libtool library 'libasn1compiler.la'
	/usr/share/automake-1.14/am/ltlibrary.am: warning: 'libasn1fix.la': linking libtool libraries using a non-POSIX
	/usr/share/automake-1.14/am/ltlibrary.am: archiver requires 'AM_PROG_AR' in 'configure.ac'
	libasn1fix/Makefile.am:6:   while processing Libtool library 'libasn1fix.la'
	/usr/share/automake-1.14/am/ltlibrary.am: warning: 'libasn1parser.la': linking libtool libraries using a non-POSIX
	/usr/share/automake-1.14/am/ltlibrary.am: archiver requires 'AM_PROG_AR' in 'configure.ac'
	libasn1parser/Makefile.am:6:   while processing Libtool library 'libasn1parser.la'
	/usr/share/automake-1.14/am/ltlibrary.am: warning: 'libasn1print.la': linking libtool libraries using a non-POSIX
	/usr/share/automake-1.14/am/ltlibrary.am: archiver requires 'AM_PROG_AR' in 'configure.ac'
	libasn1print/Makefile.am:7:   while processing Libtool library 'libasn1print.la'
	/usr/share/automake-1.14/am/ltlibrary.am: warning: 'libasn1cskeletons.la': linking libtool libraries using a non-POSIX
	/usr/share/automake-1.14/am/ltlibrary.am: archiver requires 'AM_PROG_AR' in 'configure.ac'
	skeletons/Makefile.am:21:   while processing Libtool library 'libasn1cskeletons.la'
	(...)

dunn:
I get this on Mac OS 10.10.4 too: https://gist.github.com/dunn/5fe1aae3aabf68190f27#file-01-autoreconf-L42

[...]

PR #51, opened 2014-11-06
AM_PROG_AR is missing and makes libtool ac failing on CYGWIN_NT-6.1 idefix 1.7.32(0.274/5/3) 2014-08-13 23:06 x86_64 Cygwin

zosrothko:
Signed-off-by: zosrothko <zosrothko@orange.fr>

PR #63, opened 2015-08-01
Update configure.ac

brchiu:
fixing problem found when invoke autoreconf -iv where autoreconf is of version 2.69


anatolyborodin:
Can #62 be related to this change?

anatolyborodin:
@vlm #51 is also AM_PROG_AR-related.

simo5:
This patch fixes autoreconf also on Fedora with automake 1.15

DevNaga:
same problem on my ubuntu 14.04 LTS too.. Your patch fixes it .. i tried adding directly the AM_PROG_AR to configure.ac and it worked too.

PR #80, opened 2016-02-03
Fix asn1c options dump in .c/.h preamble.

### LINKED ISSUE #55: Output files list all but the last flag in their comments
At the top of each .c and .h file generated by asn1c, there is a commented line like `asn1c -fcompound-names`, ostensibly showing the options used to generate them. This information is slightly misleading, however, since the last option is always missing; running `asn1c -fcompound-names -fno-include-deps *.asn` will produce a comment in each file reading `asn1c -fcompound-names`.

mkhon:
This fixes #55


vlm:
Fixed, thank you!

PR #81, opened 2016-02-05
Autotools glue for .l and .y

mkhon:
.l and .y are not rebuilt if changed


vlm:
I reject this patch, because this autorebuild is too sensitive wrt. to the lexx/yacc versions, and produce unanticipated build problems on some systems.

PR #82, opened 2016-02-05
Bison 3 no longer supports YYPARSE_PARAM and YYPARSE_PARAM_TYPE.

mkhon:
byacc also supports %param-type

### DIFF
--- a/libasn1parser/asn1p_y.y
+++ b/libasn1parser/asn1p_y.y
@@ -1,3 +1,5 @@
+%parse-param { void **param }
+
 %{
 
 #include <stdlib.h>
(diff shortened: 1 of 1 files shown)


vlm:
Is there a way to support both bison 2 and bison 3?

velichkov:
Hi @vlm,

> Is there a way to support both bison 2 and bison 3?

For me this fix works well with bison 2.5 and 3.0.2. Only one compiler warning needs to be fixed

```
  CC       asn1p_y.lo
asn1p_y.y: In function 'asn1p_error':
asn1p_y.y:2465:16: error: unused parameter 'param' [-Werror=unused-parameter]
 yyerror(void **param, const char *msg) {
```

I workaround it with assert in mouse07410#3 and mouse07410#5

```diff
--- a/libasn1parser/asn1p_y.y
+++ b/libasn1parser/asn1p_y.y
@@ -2463,6 +2463,7 @@ _fixup_anonymous_identifier(asn1p_expr_t *expr) {
 
 int
 yyerror(void **param, const char *msg) {
+       assert(param);
        extern char *asn1p_text;
        fprintf(stderr,
                "ASN.1 grammar parse error "
```

vlm:
Re workaround: such warnings are typically fixed by casting to void: `(void)param;` or by using `__attribute__((unused))` or the app-specific macro wrapping it, such as `GCC_UNUSED` for asn1c.

PR #83, opened 2016-03-07
fix installation of standard-modules

lukem:
Fix a typo in the definition of smodsdir so that
the standard-modules directory gets installed.


vlm:
Fixed, thank you!

PR #87, opened 2016-03-14
Delete unnecessary checks before some function calls

elfring:
- [The function "free"](http://stackoverflow.com/questions/18775608/free-a-null-pointer-anyway-or-check-first) is documented in the way that no action shall occur for a passed null pointer. Redundant safety checks can be avoided.
- The following functions perform also input parameter validation.
  - [asn1p_constraint_free](https://github.com/vlm/asn1c/blob/5c541f119e188d4ada807cc53938fff8a84c88a4/libasn1parser/asn1p_constr.c#L22)
  - [asn1p_paramlist_free](https://github.com/vlm/asn1c/blob/d8b8364c8acea078ff098d8b6fbaa0a29da9346a/libasn1parser/asn1p_param.c#L24)
  - [asn1p_ref_free](https://github.com/vlm/asn1c/blob/d8b8364c8acea078ff098d8b6fbaa0a29da9346a/libasn1parser/asn1p_ref.c#L24)
  - [asn1p_value_free](https://github.com/vlm/asn1c/blob/a9532f4d2bad200422ffacae45342c73f7cb40cb/libasn1parser/asn1p_value.c#L221)
  - [asn1p_wsyntx_free](https://github.com/vlm/asn1c/blob/c46b7cb50041897e2f6d617bdf23582d9462d4e5/libasn1parser/asn1p_class.c#L124)
  
  It is therefore not needed that a function caller repeats a corresponding check.

### DIFF
--- a/libasn1fix/asn1fix_crange.c
+++ b/libasn1fix/asn1fix_crange.c
@@ -673,10 +673,8 @@ _range_canonicalize(asn1cnst_range_t *range) {
 			range->right = tmp;
 		}
 
-		if(range->elements) {
-			free(range->elements);
-			range->elements = 0;
-		}
+		free(range->elements);
+		range->elements = 0;
 		range->el_size = 0;
 		return 0;
 	}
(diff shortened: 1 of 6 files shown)


vlm:
Fixed, thank you!

PR #89, opened 2016-05-14
Unify the naming of asn_OCTET_STRING_specifics_t variables

brchiu:
Unify the naming of asn_OCTET_STRING_specifics_t with code-generated '_specifics_t' variables to have 'asn_SPC_' prefix instead of 'asn_DEF_' prefix.

### DIFF
--- a/skeletons/ANY.c
+++ b/skeletons/ANY.c
@@ -6,7 +6,7 @@
 #include <ANY.h>
 #include <errno.h>
 
-static asn_OCTET_STRING_specifics_t asn_DEF_ANY_specs = {
+static asn_OCTET_STRING_specifics_t asn_SPC_ANY_specs = {
 	sizeof(ANY_t),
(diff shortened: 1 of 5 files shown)


mouse07410:
@brchiu I've applied your PR to [my fork](url), but am a bit at loss as to why it was needed. It doesn't seem to affect the functionality at all - so who cares what the internal names are?

brchiu:
hi, @mouse07410 , this PR itself did not fix any existing problem, however, by using consistent name across basic types, we can do further improvement to asn1c. For example, to extract various of encoder/decoder data structure of basic types to an array of function pointers. Even more, to use enum field to access this array of decoder/encoder function rather than embed the whole data structure into every new types. This could reduce the code size of final encoder/decoder in our application.

vlm:
Merged, thank you!

PR #100, opened 2016-07-07
Update 112-param-class-OK.asn1

zosrothko:
IMO, &valueField3 is a fixed-type value of type INTEGER, and thus it cannot have INTEGER:ValueSet as value.


vlm:
That's a constraint, it can be a set of numbers.

Issue #47, opened 2014-08-25
Failed X.509 example


DesmondHayes:
Tried examples/sample.source.PKIX1, but failed.
Compiled just fine with usual 'make'. Then tried to use 'x509dump' executable with 'sample-Certificate-1.der' as argument, which gave me an error 'sample-Certificate-1.der: Decode failed past byte 4: Input processing error'. Also tried other certificates, self-signed and real (from the cache of Firefox). Examples from 'http://csrc.nist.gov/groups/ST/crypto_apps_infra/documents/pkixtools/' worked OK. 'Microsoft Internet Authority' certificate from the cache of Firefox forked wine, but several others failed. Self-signed certificates generated with OpenSSL and with IIS7 failed.

daa:
adding ``-fwide-types`` to asn1c command solves this problem for me

PR #102, opened 2016-07-08
Use wide types to decode x509 certificate in examples, solves #47

daa:
Certificate serial number in `examples/sample.source.PKIX1/sample-Certificate-1.der` as in almost all real-world certificates is large number that cannot fit into native integer type, so original `x509dump` fails with this certificate. Using `-fwide-types` option for `asn1c` solves this problem and given certificate is successfully parsed.


vlm:
Thanks, merged! I forgot to do this when I changed the integer size defaults from wide to native.

PR #104, opened 2016-07-18
Fix segmentation fault in asn1c_make_identifier 

velichkov:
See http://stackoverflow.com/questions/25390720/va-arg-gives-something-strange-in-cygwin-x64

I've also fix execution of  asn1c tests. On RedHat Enterprise Linux 7 they were failing with 

```
 ../../config/test-driver: invalid option: '-C'
```

### DIFF
--- a/asn1c/tests/check-assembly.sh
+++ b/asn1c/tests/check-assembly.sh
@@ -17,7 +17,7 @@ abs_top_srcdir="${abs_top_srcdir:-$(pwd)/../../}"
 abs_top_builddir="${abs_top_builddir:-$(pwd)/../../}"
 
 if echo "$*" | grep -q -- -- ; then
-    TEST_DRIVER=$(echo "$*"  | sed -e 's/ -- .*/--/g')
+    TEST_DRIVER=$(echo "$*"  | sed -e 's/ -- .*/ -- /g')
     source_full=$(echo "$*"  | sed -e 's/.* //g')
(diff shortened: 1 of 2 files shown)


alexmurray:
I also just encountered this when using asn1c on a 64-bit platform (ie where `sizeof(int) < sizeof(char *)` - I wonder would the use of the explicit `NULL` be a better choice than `0` or the proposed `(char *)0` since that is what NULL is for and it is clear what it means semantically.

velichkov:
Hi @alexmurray 

I choose `(char *)0` mainly because in asn1c_make_identifier the va_arg is always called with `(char *)`
```c
	va_start(ap, expr);
	while((str = va_arg(ap, char *)))
```
Also in C the macro NULL is defined as an implementation-defined null pointer constant, so some compilers could just define it like ```#define NULL 0``` and not as ```(void*)0```

From [GCC manual](https://www.gnu.org/software/libc/manual/html_node/Null-Pointer-Constant.html)
> If you use the null pointer constant as a function argument, then for complete portability you should make sure that the function has a prototype declaration. Otherwise, if the target machine has two different pointer representations, the compiler won’t know which representation to use for that argument. You can avoid the problem by explicitly casting the constant to the proper pointer type, but we recommend instead adding a prototype for the function you are calling.

vlm:
Merged, thank you!

PR #109, opened 2016-07-30
Locate AR using standard macro

theirix:
- Use AM_PROG_AR to locate AR (debian complains about AC_PATH_PROG)
- Added a fallback for automake < 1.12
- Moved LT_INIT because it depends on AM_\* macro


vlm:
Merged, thank you!

PR #127, opened 2016-11-09
per encoding allow larger constraints

persandstrom:
Constraint values might be larger than what can be held in a 32bit integer. 
On example is TimestampIts defined by ETSI in http://forge.etsi.org/websvn/filedetails.php?repname=LIBS.LibIts&path=%2Ftrunk%2Fasn1%2FITS-Container%2FITS-Container.asn


persandstrom:
This solution does not work as intended

PR #132, opened 2016-12-13
Rewrite the 64bit test. Fixes #131

### LINKED ISSUE #131: autoreconf: recursion limit of 1024 exceeded
After cloning the git repo, I'm trying to build, but it fails on the first step:

```
$ autoreconf -iv
autoreconf: Entering directory `.'
autoreconf: configure.ac: not using Gettext
autoreconf: running: aclocal -I m4
/usr/bin/m4:configure.ac:107: recursion limit of 1024 exceeded, use -L<N> to change it
autom4te: /usr/bin/m4 failed with exit status: 1
aclocal: autom4te failed with exit status: 1
autoreconf: aclocal failed with exit status: 1

```

I don't know enough about autoconf to understand how to troubleshoot this. The error message suggests increasing the recursion limit to m4, but I can't see how one passes this option in. In any case 1024 seems pretty deep, maybe something else is wrong...

I'm using Centos 6, which has autoconf 2.63 and automake 1.11.1.

velichkov:
The old test was not working at all and was failing on both 32 and 64
bit systems with

```bash
configure:13545: checking size of long is longer than 32 bit
configure:13561: gcc -c -g -O2 -Wall -Wcast-qual -Wchar-subscripts
-Wmissing-prototypes -Wmissing-declarations -Wno-error=cast-align
conftest.c >&5
conftest.c: In function 'main':
conftest.c:32:34: error: 'dummy' undeclared (first use in this function)
 switch (0) case 0: case (sizeof (dummy) >= 8):;
                                  ^
conftest.c:32:34: note: each undeclared identifier is reported only once
for each function it appears in
```


vlm:
Merged (slightly altered), thank you!

PR #129, opened 2016-11-14
per encoding allow larger constraints

persandstrom:
Constraint values might be larger than what can be held in a 32bit integer.
One example is TimestampIts defined by ETSI in http://forge.etsi.org/websvn/filedetails.php?repname=LIBS.LibIts&path=%2Ftrunk%2Fasn1%2FITS-Container%2FITS-Container.asn

PR #143, opened 2017-03-17
Move type operations to new structure

brchiu:
Every type has free, print, check_constraints, ber_decoder, der_encoder, xer_decoder, xer_encoder, uper_decoder, uper_encoder and outmost_tag operations. We move them out to a separate structure asn_TYPE_operation_t.

Combined with previous logic simplification, these operations are based on ASN.1 basic types, constructed types and string types. So we can reduce the space occupied by asn_TYPE_descriptor_t variables.

With this PR, the program size of executable rrc-dump is reduced from  6983632 bytes to 6092584 bytes on my platform, roughly a 12% reduction.

Moreover, we can easily add conditional compilation for enabling or disabling various encoders and decoders, which can reduce more program size. For example, an eNB program might only require un-aligned-PER for RRC and aligned-PER for S1AP, we can remove XER and BER encoder/decoder and also their dictionaries.

It is important for embedded device which has limited memory and storage.

The capability to enable or disable different encoder and decoder is planned in next pull request.


mouse07410:
I'd love to apply your PR - but it seems to conflict with the other PRs already applied to my fork (and I don't have time now to "massage" it by hand to make it work).

Would you be willing to modify it for https://github.com/mouse07410/asn1c.git master branch?

brchiu:
Hi, @mouse07410 , thanks for your interest in this PR.

I will try send a similar PR to your repository because I also need those features you merged before, like APER, Information Object  .... etc. Thanks for your previous effort. 
However, It might take some time, please be patient.

By the way, I only verify these modification through basic 'make check' tests and I don't have real world samples to extensively verdict its correctness. If anyone can do more test on my PR is welcomed.

brchiu:
Hi, @mouse07410, 

I have created a branch on my repository [pr143_for_mouse07410](https://github.com/brchiu/asn1c/tree/pr143_for_mouse07410) which sync your master branch to my [pull 143](https://github.com/vlm/asn1c/pull/143). However, when I create a pull request to your [https://github.com/mouse07410/asn1c.git](https://github.com/mouse07410/asn1c.git), it still complain not able to auto-merge them.

It did not show which file has conflict. Any advice to proceed ?

Or I just create the pull request to your repository and let us check which file need to be edited further ?

mouse07410:
> It did not show which file has conflict. Any advice to proceed ?

I suggest cloning my repo and trying to apply your PR as a patch to it. This will show all the incompatibilities.

> Or I just create the pull request to your repository and let us check which file need to be edited further?

You can - but in my opinion the above recommendation (cloning my repo and trying your PR as a patch) would be both better and quicker.

P.S. It would be nice if you could do the above to two branches: `master` and `115`. Thanks!

P.P.S. You can do the above with other PRs that you updated (and that haven't been applied yet to my fork).

brchiu:
I indeed clone your master branch, then the comparison and merge by meld tool and hand, that's why I wonder it does not work. Anyway, I will try to let git do this merge job and let git tell me which conflict it has.

By the way, I think you already merged 115 on your master branch, don't you ? Because I saw APER related code in your master branch, however I haven't check your 115 branch in detail yet.

skeletons/INTEGER.h
```diff
@@ -43,6 +43,8 @@ xer_type_encoder_f INTEGER_encode_xer;
 per_type_decoder_f INTEGER_decode_uper;
 per_type_encoder_f INTEGER_encode_uper;
 
+#define INTEGER_constraint asn_generic_no_constraint
```
mouse07410: 1. I don't think it's a good idea.
2. Somehow my fork seems to compile and run PKIX1 sample just fine without this change.

mouse07410:
> I indeed clone your master branch, then the comparison and merge by meld tool and hand, that's why I wonder it does not work. 

If you had to do any manual editing (aka "by hand") - that's what needs to be fixed. If you just did "patch <whatever>" and it ran smoothly without any rejects - then something's strange and wrong.

> Anyway, I will try to let git do this merge job and let git tell me which conflict it has.

Yes please. I cannot accept any PR that has any conflict (or that fails the CI).

> By the way, I think you already merged 115 on your master branch, don't you ?

Ah, not quite. There are two different implementations of APER support. I maintain both, but master can have only one of them (obviously). 115 is the other one.

skeletons/INTEGER.h
brchiu: I know use `#define INTEGER_constraint asn_generic_no_constraint` might not be a good idea, but it looks like the original generated code works in this way.

I added this line because I found that the following definitions of `rfc3280-PKIX1Explicit88.asn1`  generate Version.c and CertificateSerialNumber.c and they use `INTEGER_constraint` without definition.

```
Version  ::=  INTEGER  {  v1(0), v2(1), v3(2)  }
CertificateSerialNumber  ::=  INTEGER
```

Then I check original generated Version.c and CertificateSerialNumber.c, there asn_constr_check_f is 

```
int
Version_constraint(asn_TYPE_descriptor_t *td, const void *sptr,
			asn_app_constraint_failed_f *ctfailcb, void *app_key) {
	/* Replace with underlying type checker */
	td->check_constraints = asn_DEF_INTEGER.check_constraints;
	return td->check_constraints(td, sptr, ctfailcb, app_key);
}
int
CertificateSerialNumber_constraint(asn_TYPE_descriptor_t *td, const void *sptr,
			asn_app_constraint_failed_f *ctfailcb, void *app_key) {
	/* Replace with underlying type checker */
	td->check_constraints = asn_DEF_INTEGER.check_constraints;
	return td->check_constraints(td, sptr, ctfailcb, app_key);
}
```
And in INTEGER.c, the asn_constr_check_f field of asn_DEF_INTEGER is indeed asn_generic_no_constraint.  

I did not change anything substantially but only simplified the logic of searching asn_constr_check_f field of asn_DEF_Version and asn_DEF_CertificateSerialNumber to eliminate `Version_constraint` and `CertificateSerialNumber_constraint'.

I think the question becomes "whether, at least for the two ASN.1 expressions, shall they generate corresponding constraint check functions for them or not ?" 

However, it is beyond my original purpose of this serial pull requests.

brchiu: Take another expression in `rfc3280-PKIX1Explicit88.asn1` as an example :

```
ExtensionAttribute ::=  SEQUENCE {
   extension-attribute-type [0] IMPLICIT INTEGER
                   (0..ub-extension-attributes),
   extension-attribute-value [1]
                   ANY DEFINED BY extension-attribute-type }
```
`extension-attribute-type` has a range constraint and my PR indeed generate the corresponding `memb_extension_attribute_type_constraint_1` as original code did.

This is another evidence that my PR work the same way as original code.

mouse07410: I confess to being thoroughly confused now. I understood that your `+#define INTEGER_constraint asn_generic_no_constraint` was needed because the `sample.source.PKIX1` failed to work for you without it? And I showed that it was not the case as that sample code worked for me.

???

brchiu: I am also confused about your statement "I showed that it was not the case as that sample code worked for me."

Do you mean you use (1) your own repository to generate codes for sample.source.PKIX1 or (2) use my #143 to generate code for sample.source.PKIX1 ? 

Actually if you compare the generated code using your own master repository and the generated code using my #143, you can see they are different.

Please remember to execute **make regen** to re-generate these C files and H files, otherwise, the existed files will be the same. Of course you can delete them by hand.

mouse07410: > Do you mean you use (1) your own repository to generate codes for sample.source.PKIX1 ?

Yes that's exactly what I meant. And I posted the result of running the freshly-generated `x509dump` on the provided/included `sample-Certificate-1.der`.

> Please remember to execute `make regen` to re-generate these C files and H files, otherwise, the existed files will be the same

For me `make distclean` does the job. :-)
 (Unless I'm missing something?).

brchiu: Of course your repository do not need this line, as lone as you have apply #143 to it.

Please please take a look at the generated Version.c using your repository and also the Version.c generated by my #143.

You can see the Version.c generated by your repository

```
int
Version_constraint(asn_TYPE_descriptor_t *td, const void *sptr,
			asn_app_constraint_failed_f *ctfailcb, void *app_key) {
	/* Replace with underlying type checker */
	td->check_constraints = asn_DEF_INTEGER.check_constraints;
	return td->check_constraints(td, sptr, ctfailcb, app_key);
}
......
asn_TYPE_descriptor_t asn_DEF_Version = {
	"Version",
	"Version",
	Version_free,
	Version_print,
	Version_constraint,
	Version_decode_ber,
	Version_encode_der,
	Version_decode_xer,
	Version_encode_xer,
	0, 0,	/* No PER support, use "-gen-PER" to enable */
	0, 0,	/* No APER */
	0,	/* Use generic outmost tag fetcher */
	asn_DEF_Version_tags_1,
	sizeof(asn_DEF_Version_tags_1)
		/sizeof(asn_DEF_Version_tags_1[0]), /* 1 */
	asn_DEF_Version_tags_1,	/* Same as above */
	sizeof(asn_DEF_Version_tags_1)
		/sizeof(asn_DEF_Version_tags_1[0]), /* 1 */
	0,	/* No PER visible constraints */
	0, 0,	/* Defined elsewhere */
	0	/* No specifics */
};
```
Since the `asn_constr_check_f` is indirectly called to asn_DEF_INTEGER.check_constraints, so you do not encounter compilation error I had if `#define INTEGER_constraint asn_generic_no_constraint` not added.

So your code work fine is very reasonable, and my #143 work too because I just simplify these operation functions finding logic, so it is also reasonable. 

And the remain question are, 

(1) Why asn_DEF_INTEGER and asn_DEF_NativeInteger has asn_generic_no_constraint as their asn_constr_check_f fields ?

(2) Why the following ASN.1 statement does not generate constraint check function ?'
```
Version  ::=  INTEGER  {  v1(0), v2(1), v3(2)  }
```

Honestly, I don't know. 
And it is beyond my purpose of #143, I don't plan to tackle it here.

mouse07410: > So your code works fine . . . is very reasonable, and my #143 works too

In that case and at this time I'd rather avoid a simplification whose consequences I don't fully understand.

> (1) Why `asn_DEF_INTEGER` and `asn_DEF_NativeInteger` has `asn_generic_no_constraint` as their `asn_constr_check_f` field ?

Good question. We should hope to find an answer eventually.

> (2) Why the following ASN.1 statement does not generate constraint check function ?
```
Version  ::=  INTEGER  {  v1(0), v2(1), v3(2)  }
```

This is a good question that I think we must find an answer to, and preferably sooner rather than later. It seems related to the issue described in https://github.com/vlm/asn1c/pull/129#issuecomment-286685144.

brchiu: Just put a record in case anyone is interested in this discussion. Please refer to [here](https://github.com/mouse07410/asn1c/pull/2#issuecomment-288065551)

mouse07410:
@brchiu all the work in my work should be on the branch 115. it will become the new master soon. The current master will become the new branch 129 (I'm not sure how well I'd be able to maintain that branch).

Please plan accordingly, if you'd like to see your #143 to continue in the new master of my fork.

brchiu:
Hi, @mouse07410 , 

I will prepare another patch to your new master.

PR #145, opened 2017-04-11
Skip last comma when there are no extensions

johvik:
This is to get rid of the GCC warning:
"warning: comma at end of enumerator list [-Wpedantic]"

### DIFF
--- a/libasn1compiler/asn1c_C.c
+++ b/libasn1compiler/asn1c_C.c
@@ -938,6 +938,7 @@ asn1c_lang_C_type_CHOICE(arg_t *arg) {
 		OUT("_PR_NOTHING,\t/* No components present */\n");
 		TQ_FOR(v, &(expr->members), next) {
 			if(skipComma) skipComma = 0;
+			else if (v->expr_type == A1TC_EXTENSIBLE && !TQ_NEXT(v, next)) OUT("\n");
 			else OUT(",\n");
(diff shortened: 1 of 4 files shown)


libasn1compiler/asn1c_C.c
```diff
@@ -938,6 +938,7 @@ asn1c_lang_C_type_CHOICE(arg_t *arg) {
 		OUT("_PR_NOTHING,\t/* No components present */\n");
 		TQ_FOR(v, &(expr->members), next) {
 			if(skipComma) skipComma = 0;
+			else if (v->expr_type == A1TC_EXTENSIBLE && !TQ_NEXT(v, next)) OUT("\n");
```
vlm: Why no comma only if extensible? I would assume that either `v->expr_type == A1TC_EXTENSIBLE` should not be there at all, or it should be inverted into `v->expr_type != A1TC_EXTENSIBLE`. What do you think?

johvik: I think it is needed since the A1TC_EXTENSIBLE is only a placeholder for the ellipsis and does not add any values to the enum. Note the `&& !TQ_NEXT(v, next)` which essentially checks if another value will be added.

I guess it would be possible to use other conditions if the code below was changed to only print the comment `OUT("/* Extensions may appear below */\n");` when there are extensions, if that is preferable.

vlm:
Merged, thank you!

Issue #141, opened 2017-03-03
segfault on reusing a structure after ASN_STRUCT_FREE_CONTENTS_ONLY().


dbaarda:
I'm not sure if this is a bug or working as intended. I didn't encounter this problem with v0.9.24, and am seeing it now with 0.9.27, but maybe I just got lucky before.  I also switched from native-types to wide-types which may have contributed.

When trying to reuse a message structure after using and then "emptying" it, setting attributes using things like OCTET_STRING_fromBuf() can segfault. An example LDAP code fragment looks like this;

```C
msg = calloc(1,sizeof(LDAPMessage_t))
msg->messageID = msgid;
res->protocolOp.present = LDAPMessage__protocolOp_PR_searchResEntry;
SearchResultEntry_t *searchResEntry = &res->protocolOp.choice.searchResEntry;
/* Populate the msg searchResEntry attributes here... */
...
/* Now empty the msg and reuse it for a searchResDone response. */
ASN_STRUCT_FREE_CONTENTS_ONLY(asn_DEF_LDAPMessage, msg);
msg->protocolOp.present = LDAPMessage__protocolOp_PR_searchResDone;
SearchResultDone_t *searchResDone = &res->protocolOp.choice.searchResDone;
/* And now this will segfault... */
OCTET_STRING_fromString(&searchResDone->matchedDN, basedn);
```

The problem is that OCTET_STRING_fromString() tries to free the existing matchedDN attribute before alloc/set the new one. However, the existing matchedDN attribute contains a garbage pointer left over from the previous searchResEntry message.

I realize this is pretty normal C behavior; freeing things doesn't re-initialize/zero the pointers, and it's up to the caller to do that. However, there appears to be some effort in things like OCTET_STRING_free() to leave the free'd structure in a valid initialized state, doing things like zeroing the pointer after free. Digging around other *_free() functions it looks like most others don't do this.

I think the right thing here is to adhere to C conventions and have ASN_STRUCT_FREE_CONTENTS_ONLY() offer no guarantees that the structure is left in a valid initialized/zeroed state. In that case I think *_free() functions should never expend effort on things like zeroing pointers after free(), because that just masks that this cannot be relied on.

mouse07410:
[...]

PR #93, opened 2016-05-16
Simplify finding decoder logic

brchiu:
Taking 'FileCreationTimeStamp' of sample TAP3 as example, the corresponding ASN.1 expression of tap3.asn1 are :

```
FileCreationTimeStamp ::= [APPLICATION 108] DateTimeLong

DateTimeLong ::= SEQUENCE 
{
    localTimeStamp     LocalTimeStamp OPTIONAL, -- *m.m.
    utcTimeOffset      UtcTimeOffset  OPTIONAL
}
```

Part of generated FileCreationTimeStamp.c is : 

```
...
asn_dec_rval_t
FileCreationTimeStamp_decode_ber(asn_codec_ctx_t *opt_codec_ctx, asn_TYPE_descriptor_t *td,
        void **structure, const void *bufptr, size_t size, int tag_mode) {
    FileCreationTimeStamp_1_inherit_TYPE_descriptor(td);
    return td->ber_decoder(opt_codec_ctx, td, structure, bufptr, size, tag_mode);
}

static void
FileCreationTimeStamp_1_inherit_TYPE_descriptor(asn_TYPE_descriptor_t *td) {
    td->free_struct    = asn_DEF_DateTimeLong.free_struct;
    td->print_struct   = asn_DEF_DateTimeLong.print_struct;
    td->check_constraints = asn_DEF_DateTimeLong.check_constraints;
    td->ber_decoder    = asn_DEF_DateTimeLong.ber_decoder;
    td->der_encoder    = asn_DEF_DateTimeLong.der_encoder;
    td->xer_decoder    = asn_DEF_DateTimeLong.xer_decoder;
    td->xer_encoder    = asn_DEF_DateTimeLong.xer_encoder;
    td->uper_decoder   = asn_DEF_DateTimeLong.uper_decoder;
    td->uper_encoder   = asn_DEF_DateTimeLong.uper_encoder;
    if(!td->per_constraints)
        td->per_constraints = asn_DEF_DateTimeLong.per_constraints;
    td->elements       = asn_DEF_DateTimeLong.elements;
    td->elements_count = asn_DEF_DateTimeLong.elements_count;
    td->specifics      = asn_DEF_DateTimeLong.specifics;
}

asn_TYPE_descriptor_t asn_DEF_FileCreationTimeStamp = {
    "FileCreationTimeStamp",
    "FileCreationTimeStamp",
    FileCreationTimeStamp_free,
    FileCreationTimeStamp_print,
    FileCreationTimeStamp_constraint,
[...]

PR #146, opened 2017-04-11
Zero out memory when free content only

brchiu:
When calling SET_free(), SET_OF_free(), SEQUENCE_free(), CHOICE_free() and OCTET_STRING_free() with contents_only = 1, we zero out the memory pointed by ptr for struct_size bytes.

This is trying to solve Issue #141.

However, since #93 has not been accepted yet, for derived type which inherit from others and has null 'specifics' field might still has #141 issue.

### DIFF
--- a/skeletons/OCTET_STRING.c
+++ b/skeletons/OCTET_STRING.c
@@ -1746,6 +1746,8 @@ OCTET_STRING_free(asn_TYPE_descriptor_t *td, void *sptr, int contents_only) {
 
 	if(!contents_only) {
 		FREEMEM(st);
+	} else if(td->specifics) {
+		memset(st, 0, ((asn_OCTET_STRING_specifics_t *)(td->specifics))->struct_size);
 	}
 }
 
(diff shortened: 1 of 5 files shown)


mouse07410:
@brchiu I like this in general. But my concern is - this PR seems to be doing more than should be needed to mitigate the problem (I'd like to hear from @dbaarda as well):

- The problem seems to be that in case of a union, attempts to re-use pre-allocated structures hit wrong elements (e.g., where there was something else, now may be a placeholder for an `OCTET STRING`).
- Thus to remedy, the union types should be zeroed out so the pre-allocated memory cannot be mis-interpreted.

But I don't see such possibility of mis-interpretation with `SEQUENCE`, `OCTET STRING`, and probably `SET OF` (not sure here). So why do we want to zero those containers?

vlm:
I was reluctant to merge this patch, because it introduces recursive (read: quadratic) memory resetting behavior when contents_only option is present. Consider the hierarchical structures. Freeing them would memset() each of the inner structures, then memset() the outer structure, and so on. This essentially resets each byte of the structure as many times as the number of layers in the hierarchy.

Therefore I introduced `ASN_STRUCT_RESET()` macro which does clear the contents, and discouraged using `ASN_STRUCT_FREE_CONTENTS_ONLY()` in the application code.

https://github.com/vlm/asn1c/commit/8d99d7b5c87d5a805472a51007d5f640932d435e

vlm:
Closing as a substitute implementation is available.

PR #126, opened 2016-11-08
Fix many memory leaks reported by valgrind

brchiu:
Using Valgrind to check memory leakage during compiling rrc-7.1.0.asn1 :

```
==32511== LEAK SUMMARY:
==32511==    definitely lost: 664,203 bytes in 6,379 blocks
==32511==    indirectly lost: 107,661,853 bytes in 2,273,838 blocks
==32511==      possibly lost: 256 bytes in 4 blocks
==32511==    still reachable: 86,691 bytes in 98 blocks
==32511==         suppressed: 0 bytes in 0 blocks
==32511== 
==32511== ERROR SUMMARY: 148 errors from 148 contexts (suppressed: 0 from 0)
==32511== ERROR SUMMARY: 148 errors from 148 contexts (suppressed: 0 from 0)
```

Many of them are fixed with this pull request, but there exist several leaks I am not able to solve. 
[Detail of remaining leaks when compiling rrc-7.1.0.asn1](https://gist.github.com/brchiu/589126422c98e568fab676ddfe0ea45a)

```
==32093== LEAK SUMMARY:
==32093==    definitely lost: 259,209 bytes in 2,478 blocks
==32093==    indirectly lost: 168,787 bytes in 1,321 blocks
==32093==      possibly lost: 0 bytes in 0 blocks
==32093==    still reachable: 82,185 bytes in 6 blocks
==32093==         suppressed: 0 bytes in 0 blocks
==32093== 
==32093== ERROR SUMMARY: 8 errors from 8 contexts (suppressed: 0 from 0)
==32093== ERROR SUMMARY: 8 errors from 8 contexts (suppressed: 0 from 0)
```

PR #153, opened 2017-05-05
Fix some memory leakage found till now

brchiu:
Memory allocated in asn1p_l.c and asn1p_y.c are freed.

Remaining two leaked memory blocks are allocated inside function asn1f_printable_value() and asn1c_make_identifier() hold by static variables. Please refer to [Valgrind log.](https://gist.github.com/brchiu/d0ef9f32e373095a9703712aa2b4e9c9)

asn.1 modules under `examples` directories and some more asn.1, ex. LTE-RRC, S1AP, RANAP, DSRC, NBAP are used for verification.

When the parser ends with unexpected token or grammar error, memory allocated by `asn1_module_new()`.... etc are not able freed because `asn1p_parse()` does not return a valid `asn1p_t *` pointer to caller. Perhaps adding a suitable hook to call `asn1p_free()` inside parser is needed.

Of course, there might still be un-found leakages exist in other asn,1 syntax cases.

### DIFF
--- a/libasn1fix/asn1fix_constr.c
+++ b/libasn1fix/asn1fix_constr.c
@@ -60,7 +60,7 @@ asn1f_pull_components_of(arg_t *arg) {
 		coft = asn1p_expr_clone(terminal, 1 /* Skip extensions */);
 		if(!coft) return -1;	/* ENOMEM */
 
-		if(0) {
+		if(1) {
 			asn1p_expr_free(memb);	/* Don't need it anymore*/
 		} else {
 			/* Actual removal clashes with constraints... skip. */
--- a/asn1c/asn1c.c
+++ b/asn1c/asn1c.c
@@ -65,6 +65,7 @@ main(int ac, char **av) {
     int ret;                        /* Return value from misc functions */
     int ch;                         /* Command line character */
     int i;                          /* Index in some loops */
+    int exit_code = 0;              /* Exit code */
 
     /*
      * Process command-line options.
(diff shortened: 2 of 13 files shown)


libasn1fix/asn1fix_constr.c
```diff
@@ -60,7 +60,7 @@ asn1f_pull_components_of(arg_t *arg) {
 		coft = asn1p_expr_clone(terminal, 1 /* Skip extensions */);
 		if(!coft) return -1;	/* ENOMEM */
 
-		if(0) {
+		if(1) {
```
vlm: Questionable. Will likely break something else, such as constraints, as indicated in the following comment.

brchiu: No. Currently `asn1c` does not clash because I solved the clash root cause in scenarios I have seen.

For examples : 
1. If a constraint is of type `ACT_EL_TYPE`, the `asn1p_expr_t *` it referenced might be freed twice in `asn_expr_free()` and `asn1p_constraint_free()` then crash. 
I solved it in previous [pr 126](https://github.com/vlm/asn1c/pull/126/files) by introducing a `ref_cnt` field in `asn1p_expr_t`.
2. In `asn1p_y.y`, `$$->Identifier = "?"` cause clash when trying to free a constant character variable, it is solved by change it to `$$->Identifier = strdup("?");` in this pr.

Of course, I admit that perhaps there is still other asn.1 syntax combination will cause asn1c crash that I have encountered yet. But currently the test cases and various aforementioned asn.1 module examples are safe for this modification.

By the way, because `asn1p_y.c` won't be automatically generated for your repository, in addition to `asn1p_y.y`, I manually modify `asn1p_y.c` at corresponding code segment.

vlm: Thank you for this explanation.

The asn1p_y problem is tricky, because autorecompile is very bison-specific and breaks on systems with preinstalled old bison.

PR #158, opened 2017-05-09
Fix duplicate 'const' warnings

johvik:
Since the typedef of asn_INTEGER_specifics_t also contains a const the
compiler complained about:
"warning: duplicate ‘const’ [-Wpedantic]"

### DIFF
--- a/libasn1compiler/asn1c_C.c
+++ b/libasn1compiler/asn1c_C.c
@@ -203,7 +203,7 @@ asn1c_lang_C_type_common_INTEGER(arg_t *arg) {
 			OUT("\t/* This list is extensible */\n");
 		OUT("};\n");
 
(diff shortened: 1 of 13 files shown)


mouse07410:
@johvik I'm not sure if https://github.com/mouse07410/asn1c.git fork needs this PR. Would you like to take a look, and if this patch is needed there, submit a PR against it as well?

johvik:
@mouse07410 I had a quick look at the code and it appears to be two consts on that fork. I would prefer it if someone else did the PR :wink:

libasn1compiler/asn1c_C.c
206: `OUT("const asn_INTEGER_specifics_t asn_SPC_%s_specs_%d = {\n",`
239: `OUT("const asn_INTEGER_specifics_t asn_SPC_%s_specs_%d = {\n",`

skeletons/INTEGER.h
28: `typedef const struct asn_INTEGER_specifics_s {`

Issue #124, opened 2016-10-28
Arguments passed to ASN_DEBUG are evaluated when EMIT_ASN_DEBUG != 1


mattipee:
```
#else   /* EMIT_ASN_DEBUG != 1 */
< static inline void ASN_DEBUG(const char *fmt, ...) { (void)fmt; }
> #define ASN_DEBUG(...) {}
#endif  /* EMIT_ASN_DEBUG */
```

In asn_internal.h, the definition of ASN_DEBUG as a function means all arguments passed to it will be evaluated and then ignored. By replacing with an empty #define, arguments are not evaluated.

I got a 4x speedup by avoiding unnecessary calls to ber_tlv_tag_string(), etc.

mattipee:
Revision 2b290be in response to pull request #172 closes this issue.

PR #172, opened 2017-08-03
Redefine ASN_DEBUG in asn_internal.h (EMIT_ASN_DEBUG != 1)

mattipee:
Change proposed in #124.

By evaluating the arguments passed to "static void ASN_DEBUG(...)", many unnecessary calls are made to ber_tlv_tag_string(). My experience suggests replacing the dummy function definition with an empty macro results in a significant decrease in CPU time.


vlm:
The reason the static function definition was there is because variadic macros (`ASN_DEBUG(`**...**`)`) is not a C90 conformant feature. I intentionally keep skeletons and generated code C90-conformant.

However, I see a definite value in your patch, so I rewrote it as https://github.com/vlm/asn1c/commit/2b290bef16a01beebb4bbf7212f789241d973d7e

Thank you!

mattipee:
Thank you.

Issue #122, opened 2016-10-28
-fno-constraints emits code with reference to constraint functions


mattipee:
With 0.9.27, -fno-constraints generates code that fails to compile for me. Visual inspection of master suggests the issue persists there.

The following patch in asn1c_C.c:emit_type_DEF works for me.

```
<       FUNCREF(constraint);

---
>       if(!arg->flags & A1C_NO_CONSTRAINTS)
>           FUNCREF(constraint);
>       else
>           OUT("0,\t/* No check because of -fno-constraints */\n");
```

vlm:
Applied, thanks!

PR #173, opened 2017-08-04
Omit constraints code if -fno-constraints

mattipee:
Raised in #122, prevent reference to constraints in generated code if compiled with -fno-constraints


mattipee:
Looks like this has failed due to the following...
```
asn1p_integer.c:105:20: error: comparison of integers of different signs: 'int'
      and 'unsigned long' [-Werror,-Wsign-compare]
        assert(ret < sizeof(static_buf));
               ~~~ ^ ~~~~~~~~~~~~~~~~~~
```

vlm:
Applied and fixed, thanks!

PR #176, opened 2017-08-10
Use special name decoration for embedded anonymous types

DanyaFilatov:
Additional type name decoration needs to be added when two different types definitions contains members of type sequence with the same name.
For example:

EccP256CurvePoint ::= CHOICE  {
    x-only           OCTET STRING (SIZE (32)),
    fill             NULL,  -- consistency with 1363 / X9.62
    compressed-y-0   OCTET STRING (SIZE (32)),
    compressed-y-1   OCTET STRING (SIZE (32)),
    uncompressed     SEQUENCE  {
        x OCTET STRING (SIZE (32)),
        y OCTET STRING (SIZE (32))
    }
}

EccP384CurvePoint::= CHOICE  {
    x-only          OCTET STRING (SIZE (48)),
    fill            NULL, -- consistency w 1363 / X9.62
    compressed-y-0  OCTET STRING (SIZE (48)),
    compressed-y-1  OCTET STRING (SIZE (48)),
    uncompressed    SEQUENCE  {
        x OCTET STRING (SIZE (48)),
        y OCTET STRING (SIZE (48))
    }
}

Both types has the 'uncompressed' member with implicit type specification, which generates identical structure types in output C files.


vlm:
Would you please add a test case into `./tests/` that verifies what you expect with `-P` flag.

DanyaFilatov:
@vlm I added the minimal valid asn1 file named 142-anonymous-types-deco-OK.asn1
Compilation of generated code is failed without my changes.

PR #177, opened 2017-08-13
Adding support to build with bacnet.asn1

ringlej:
Hi. I've been working on adding support to be able to build against the bacnet.asn1.

1) I had to make some adjustments to bacnet.asn1, because it contains nested `SEQUENCE OF SEQUENCE` and `SEQUENCE OF CHOICE` and asn1c generates code that does not compile with gcc in these cases. I worked around this by modifying bacnet.asn1 so that the nested elements were externalized in the asn1 code.

2) The bacnet.asn1 code had some places where it had ENUMERATED in a CHOICE, and the asn1c compiler would generate empty `typedef enum` definitions that would not compile.
```
BACnetTimerStateChangeValue ::= CHOICE {
        null                    NULL,
        boolean                 BOOLEAN,
        unsigned                Unsigned,
        integer                 INTEGER,
        real                    REAL,
        double                  Double,
        octetstring             OCTET STRING,
        characterstring         CharacterString,
        bitstring               BIT STRING,
        enumerated              ENUMERATED,
        date                    Date,
        time                    Time,
        objectidentifier        BACnetObjectIdentifier,
        no-value                [0] NULL,
        constructed-value       [1] ABSTRACT-SYNTAX.&Type,
        datetime                [2] BACnetDateTime,
        lighting-command        [3] BACnetLightingCommand
        }
```

I fixed this in libasn1compiler/asn1c_C.c so that if there are no enum members, then don't generate the `typedef enum`

3) I added -lm to LIBS, because REAL.c references `ilogb()`, which requires linking against libm

4) REAL.c and NativeReal.c are missing some `#ifndef ASN_DISABLE_PER_SUPPORT` to remove functions that call `OCTET_STRING_decode_uper` and `OCTET_STRING_encode_uper`

I can now successfully do the asn1c compile with the:
`asn1c -fcompound-names -findirect-choice -pdu=BACnetPDU ../bacnet.asn1`
and compile and link with:
`make -f Makefile.am.sample`

### DIFF
--- a/skeletons/BOOLEAN.c
+++ b/skeletons/BOOLEAN.c
@@ -415,7 +415,19 @@ BOOLEAN_random_fill(const asn_TYPE_descriptor_t *td, void **sptr,
         }
     }
 
-    /* Simulate booleans that are sloppily set */
-    *st = asn_random_between(0, 1) ? asn_random_between(1, 0x7fffffff) : 0;
+    /* Simulate booleans that are sloppily set and biased. */
+    switch(asn_random_between(0, 7)) {
+    case 0:
+    case 1:
+    case 2:
+        *st = 0; break;
+    case 3: *st = -1; break;
+    case 4: *st = 1; break;
+    case 5: *st = INT_MIN; break;
+    case 6: *st = INT_MAX; break;
+    default:
+        *st = asn_random_between(INT_MIN, INT_MAX);
+        break;
+    }
     return result_ok;
 }
--- a/skeletons/asn_random_fill.c
+++ b/skeletons/asn_random_fill.c
@@ -20,10 +20,37 @@ asn_random_fill(const struct asn_TYPE_descriptor_s *td, void **struct_ptr,
     }
 }
 
+static uintmax_t
+asn__intmax_range(intmax_t lb, intmax_t ub) {
+    assert(lb <= ub);
+    if((ub < 0) == (lb < 0)) {
+        return ub - lb;
+    } else if(lb < 0) {
+        return 1 + ((uintmax_t)ub + (uintmax_t)-(lb + 1));
+    } else {
+        assert(!"Unreachable");
+        return 0;
+    }
+}
+
 intmax_t
-asn_random_between(intmax_t a, intmax_t b) {
-    assert(a <= b);
-    assert((b-a) < RAND_MAX);
-    if(a == b) return a;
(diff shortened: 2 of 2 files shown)


ringlej:
I created issue #178 in regards to bullet 1) in the description of this PR

vlm:
Thank you for your efforts!

1. Noted. Would you please supply a minimal example of that?

2. ENUMERATED enumerations list must not be absent, be empty, or contain extension marker in the first position. This is X.680, item 20.1. So, the spec is incorrect.

3. Copied, thanks!

4. It contains these in the master, please check.

ringlej:
Example for bullet point 1 is in #178

ringlej:
Regarding bullet point 2, I'm not sure that the bacnet.asn1 is wrong in this case. This is not an ENUMERATED notation as defined in section 20 of X.680, but rather a use of ENUMERATED as the Type in a NamedType in a CHOICE (section 29)

brchiu:
@ringlej , 

I guess @vlm is asking for adding test cases into `tests/tests-asn1c-compiler` directory.

The naming rule is quite simple and it's not difficult to add it by mimic other ones. 

For example, using other test files as skeleton, you can copy your example into a file called `xxx-enum-in-choice.asn1`, where `xxx` is the next unused test numbering. 

For example, tests case with largest sequence no now is 158, so next is 159, i.e. the file name is `159-enum-in-choice.asn1`.

```
-- OK: Everything is fine

-- iso.org.dod.internet.private.enterprise (1.3.6.1.4.1)
-- .spelio.software.asn1c.test (9363.1.5.1)
-- .159

ModuleEnumInChoice
	{ iso org(3) dod(6) internet(1) private(4) enterprise(1)
		spelio(9363) software(1) asn1c(5) test(1) 159 }
DEFINITIONS ::= BEGIN

.... adding your example here !!

END
```
Then re-direct the output of `asn1c -P 159-enum-in-choice.assn1` to `159-enum-in-choice.asn1.-P`.
Put the two files : `159-enum-in-choice.assn1` and `159-enum-in-choice.asn1.-P` under `tests/tests-asn1c-compiler` then add them and commit.

By the way, a quick glance at your code, there is already a `asn1f_count_children()` to count the number of child of input expression. Perhaps you can shorten your code with it.  

Sorry I don't have code review right to your pull request, the only way is to comment your code here.

PR #181, opened 2017-08-16
Don't DER encode the default value of SEQUENCE/SET elements

### LINKED ISSUE #79: Incorrect DER output
`echo 'MAMBAQA=' | base64 -d | certlint-x509helper -p BasicConstraints -oder - | openssl asn1parse -inform DER`

The structure is defined as

``` text
BasicConstraints ::= SEQUENCE {
     cA                      BOOLEAN DEFAULT FALSE,
     pathLenConstraint       INTEGER (0..MAX) OPTIONAL }
```

According to DER

> 11.5 Set and sequence components with default value
> The encoding of a set value or sequence value shall not include an encoding for any component value which is equal to its default value.

What is observed is that the DER output contains false, which is the default value. The output in this example should be an empty sequence.

sleevi:
Per X.690 (2002), Section 11.5, for DER, set and sequence values should not
include elements equal to their default value.

This was already handled by the UPER encoder, but not by the DER
encoder.

Fixes #79


vlm:
@sleevi, thank you for this contribution! Would you please indicate that you agree to the Contributor License Agreement at http://lionet.info/asn1c/cla?

sleevi:
@vlm Done

vlm:
Thank you very much! Merged with minor changes.

PR #197, opened 2017-09-01
Added notes to J2735 and 1609.2 README files

hmusavi:
To minimize new users running into build issues, I would like to suggest that the provided notes or something to that effect be added to the J2735 and 1609.2 README files.


vlm:
asn1c now suggests `-fcompound-names` flag by automatically analyzing the name clashes.

PR #202, opened 2017-09-13
split Makefile.am.sample

DanyaFilatov:
In order to facilitate the integration of generated sources into the project tree, I would propose to split the generated Makefile.am.sample to 2 files: Makefile.am.sample containing only list of source files and Makefile.sample containing rules to generate the sample encoder/decoder.
In this case the Makefile.am.sample can be directly included by other makefiles.


vlm:
The problem with this approach is that .am files (the ones which are used by autotools) have to have sources inside them, not on the side.

Any idea how to combine these?

DanyaFilatov:
This is exactly my idea to keep the list of sources in the Makefile.am.sample (or even already rename it to the Makefile.am) to be able to use it directly in autotools or include it into other Makefiles.
Otherwise I have to edit the generated Makefile.am.sample after each regeneration.

vlm:
How about implementing `--make Makefile` or `--automake Makefile.am` options? This way, we can add `--cmake` and others, when needed.

DanyaFilatov:
Sure, it could be very good to support other build environments when it will be needed.

Actually compiler shall generate the list of source files to be integrated in some external project. The Makefile.am provides this list and the Makefile.sample is an example of such a project.

So that, Makefile.am contains list of module source files and all necessary variables and defines to build the library and can be used directly for automake or can be included in any external Makefile.
Makefile.sample includes this file and builds the converter utility.

Please have a look in last commit in the CR. I removed all converter sources from the Makefile.am

vlm:
The original generated file was named `Makefile.am.sample` and not `Makefile.am` to avoid accidental clobbering existing Makefile.am in the existing projects during recompilation.

So I reworked your patch to support splitting into:

| Makefile name | Description |
| :------------- |  :-------------- |
| `Makefile.am.example` | Defines the converter-example binary |
| `Makefile.am.libasncodecs` | Defines the libcodecs library |

Thank you for your contribution!

PR #203, opened 2017-09-13
suppress warning in x64

DanyaFilatov:
suppress warning in x64 where (unsigned) is 32 bits and pointer is 64. The size_t looks better here.

### DIFF
--- a/skeletons/converter-sample.c
+++ b/skeletons/converter-sample.c
@@ -106,7 +106,7 @@ ats_simple_name(enum asn_transfer_syntax syntax) {
 }
 
 
-#define OP_OFFSET(fname)  (unsigned)(&(((asn_TYPE_operation_t *)0)->fname))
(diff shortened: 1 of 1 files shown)


skeletons/converter-sample.c
```diff
@@ -106,7 +106,7 @@ ats_simple_name(enum asn_transfer_syntax syntax) {
 }
 
 
-#define OP_OFFSET(fname)  (unsigned)(&(((asn_TYPE_operation_t *)0)->fname))
+#define OP_OFFSET(fname)  (size_t)(&(((asn_TYPE_operation_t *)0)->fname))
```
vlm: If you want to play it even safer, use diffptr_t!

DanyaFilatov:
done. OK for ptrdiff_t. Tested with cygwin64 and MSVC.

vlm:
`unsigned codec_offset` in the following structure ought to be diffptr_t as well.

PR #216, opened 2017-10-09
gcc 7.2.1 fixes

velichkov:
- Include getopt.h instead of unistd.h
- Use unsigned max int type when converting integers

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ compiler:
   - gcc
   - clang
 before_install:
-  - sudo apt-get install -y lcov
+  - sudo apt-get install -y lcov libasan*
   - gem install coveralls-lcov
 script:
     - autoreconf -iv
--- a/skeletons/constr_SEQUENCE.c
+++ b/skeletons/constr_SEQUENCE.c
@@ -994,6 +994,9 @@ void
 SEQUENCE_free(const asn_TYPE_descriptor_t *td, void *sptr,
               enum asn_struct_free_method method) {
     size_t edx;
+    const asn_SEQUENCE_specifics_t *specs =
+        (const asn_SEQUENCE_specifics_t *)td->specifics;
+    asn_struct_ctx_t *ctx; /* Decoder context */
 
 	if(!td || !sptr)
 		return;
@@ -1012,6 +1015,10 @@ SEQUENCE_free(const asn_TYPE_descriptor_t *td, void *sptr,
 			ASN_STRUCT_FREE_CONTENTS_ONLY(*elm->type, memb_ptr);
 		}
 	}
+
+	/* Clean parsing context */
+	ctx = (asn_struct_ctx_t *)((char *)sptr + specs->ctx_offset);
+	FREEMEM(ctx->ptr);
 
     switch(method) {
     case ASFM_FREE_EVERYTHING:
--- a/skeletons/xer_encoder.c
+++ b/skeletons/xer_encoder.c
@@ -230,6 +230,8 @@ xer_equivalent(struct asn_TYPE_descriptor_s *td, void *struct1,
         return XEQ_ROUND_TRIP_FAILED;
     }
 
-    return XEQ_SUCCESS;
+	FREEMEM(xb1.buffer);
+	FREEMEM(xb2.buffer);
+	return XEQ_SUCCESS;
 }
 
--- a/skeletons/xer_encoder.c
+++ b/skeletons/xer_encoder.c
@@ -80,7 +80,9 @@ xer__buffer_append(const void *buffer, size_t size, void *app_key) {
         size_t new_size = 2 * (xb->allocated_size ? xb->allocated_size : 64);
         char *new_buf = MALLOC(new_size);
         if(!new_buf) return -1;
-        memcpy(new_buf, xb->buffer, xb->buffer_size);
+        if (xb->buffer) {
+            memcpy(new_buf, xb->buffer, xb->buffer_size);
+        }
         FREEMEM(xb->buffer);
         xb->buffer = new_buf;
         xb->allocated_size = new_size;
--- a/tests/tests-c-compiler/Makefile.am
+++ b/tests/tests-c-compiler/Makefile.am
@@ -15,6 +15,7 @@ TESTS_ENVIRONMENT= \
 				   CXXFLAGS="${CXXFLAGS}" \
 				   LDFLAGS="${LDFLAGS}" \
 				   LIBFUZZER_CFLAGS="${LIBFUZZER_CFLAGS}" \
+				   ASAN_ENV_FLAGS="@ASAN_ENV_FLAGS@"       \
 				   srcdir=${srcdir} \
 				   abs_top_srcdir=${abs_top_srcdir} \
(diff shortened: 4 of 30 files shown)


.travis.yml
```diff
@@ -7,7 +7,7 @@ before_install:
   - gem install coveralls-lcov
 script:
     - autoreconf -iv
-    - ./configure --enable-Werror --enable-code-coverage
+    - ./configure --enable-Werror --enable-code-coverage --enable-test-asan --enable-test-ubsan
```
vlm: These are automatically enabled, if available.

vlm:
Do you see ASAN_ENV_FLAGS set to something on your system (look in the generated Makefiles)? If so,

Would you mind exposing ASAN_ENV_FLAGS during the `test-c-compiler` checks by modifying `test/tests-c-compiler/check-assembly.sh` (similar to how `../tests-randomized/check-bundles.sh` is invoked) to ensure that your changes really catch memory allocation errors? In other words, I really appreciate cleaning your leaks cleanup, but it better be done under LeakSanitizer instrumentation.

velichkov:
>Do you see ASAN_ENV_FLAGS set to something on your system (look in the generated Makefiles)?

ASAN_ENV_FLAGS =  ASAN_OPTIONS=detect_leaks=1

>Would you mind exposing ASAN_ENV_FLAGS during the test-c-compiler checks by modifying test/tests-c-compiler/check-assembly.sh (similar to how ../tests-randomized/check-bundles.sh is invoked) to ensure that your changes really catch memory allocation errors? In other words, I really appreciate cleaning your leaks cleanup, but it better be done under LeakSanitizer instrumentation.

No problem, I will try to expose it.

velichkov:
> Would you mind exposing ASAN_ENV_FLAGS during the test-c-compiler checks by modifying test/tests-c-compiler/check-assembly.sh

Done. 
```
$make check
make[1]: Leaving directory '/home/vasko/sources/asn1c/tests/tests-c-compiler/test-check-03.-fwide-types'
ASAN_OPTIONS=detect_leaks=1 ./check-program
<red/> -> 0 == 0
```
> I really appreciate cleaning your leaks cleanup, but it better be done under LeakSanitizer instrumentation.

Note that I'm testing with Fedora 26 and when the ASAN_OPTIONS is not explicitly specified it seems it is enabled by default (I don't have it in my environment as well), so all of the above fixes are for errors detected by the LeakSanitizer

vlm:
I am a tad confused. Do you imply that you fixed these errors in the first place because ASAN_ENV_FLAGS were already properly exposed?

With the master, where ASAN_ENV_FLAGS configuration is the same as when you were starting your branch, my `make check` doesn't break on leaks. How had it caught leaks in your code?

In the master, ASAN_ENV_FLAGS are set, but they're not used everywhere by default, because it would break a lot at the same time. I intended to expose it gradually, as I go and fix the leaks in the code. So it should not have exposed the leaks that you fixed in this branch. So what is happening?..

velichkov:
> Do you imply that you fixed these errors in the first place because ASAN_ENV_FLAGS were already properly exposed?

No, the ASAN_ENV_FLAGS was not exposed, but the `asn1c/tests/tests-c-compiler` tests are compiled with various `-fsanitize` flags

```
$grep sanitize tests/tests-c-compiler/Makefile
SANITIZER_CFLAGS =  -fsanitize=undefined -fno-sanitize-recover=undefined -fno-omit-frame-pointer -fsanitize=address -fno-omit-frame-pointer
SKELETONS_CFLAGS =   -std=c89 -Wpedantic  -fsanitize=undefined -fno-sanitize-recover=undefined -fno-omit-frame-pointer -fsanitize=address -fno-omit-frame-pointer
TESTSUITE_CFLAGS =  -std=gnu99 -Wno-error=unused-variable  -fsanitize=undefined -fno-sanitize-recover=undefined -fno-omit-frame-pointer -fsanitize=address -fno-omit-frame-pointer
```
```
cc -I. -std=gnu99 -Wno-error=unused-variable  -fsanitize=undefined -fno-sanitize-recover=undefined -fno-omit-frame-pointer -fsanitize=address -fno-omit-frame-pointer -g -O2 -Wall -Wcast-qual -Wchar-subscripts -Wmissing-prototypes -Wmissing-declarations -Wno-error=attributes -Wno-error=cast-align -std=gnu99 -g -O1 -DSRCDIR=../. -DASN_DISABLE_OER_SUPPORT -DASN_DISABLE_PER_SUPPORT  -I. -o T.o -c T.c
```
and the check-program gets linked with asan libraries
```
$ldd check-program  | grep san
	libasan.so.4 => /lib64/libasan.so.4 (0x00007f997daf6000)
	libubsan.so.0 => /lib64/libubsan.so.0 (0x00007f997cac1000)
```
and most probably because of this the LeakSanitizer is enabled even though the ASAN_ENV_FLAGS is not exported.

> With the master, where ASAN_ENV_FLAGS configuration is the same as when you were starting your branch, my make check doesn't break on leaks. How had it caught leaks in your code?

Most probably different compiler versions with different default behaviour. I'm testing on Fedora 26 with `gcc version 7.2.1 20170915`, `glibc-2.25-10.fc26.x86_64`, `libasan-7.2.1-2.fc26.x86_64` and `libubsan-7.2.1-2.fc26.x86_64`

> In the master, ASAN_ENV_FLAGS are set, but they're not used everywhere by default, because it would break a lot at the same time. I intended to expose it gradually, as I go and fix the leaks in the code. So it should not have exposed the leaks that you fixed in this branch. So what is happening?..

I have plans to implement a Docker build using recent version of Fedora, Ubuntu or another distribution that has a recent compiler and libraries versions. This docker build could be used later in the Travis as well because currently Travis uses Ubunty Trusty which is quite old.

Are you interested in such a build and which distribution do you prefer?

vlm:
Merged (reordered/modified), thank you!

vlm:
Oh, different default behavior, nice (not really).

1. Does your compiler have ability to disable leak check by default, such as by passing "ASAN_OPTIONS=detect_leaks=0"?

2. I have no interest in one-off binary builds. I'd essentially be endorsing (and perhaps, supporting for perpetuity) someone else's outdated binaries, which is a double-wham. On the other hand, if you set up something that creates binaries automatically on commit, it would make sense to at least point out to such system for other's benefit.

velichkov:
> Does your compiler have ability to disable leak check by default, such as by passing "ASAN_OPTIONS=detect_leaks=0"?

Yes it does.

> I have no interest in one-off binary builds. I'd essentially be endorsing (and perhaps, supporting for perpetuity) someone else's outdated binaries, which is a double-wham.

I have no plans to distribute binaries as well. My idea is to have a reproducible build with known dependency versions, that's used mainly for development and testing. It would facilitates debugging issues like this one that reproduce in one environment but not in another, so I find it useful to have a known working environment(s) that is(are) easy to setup.  It would be one (or several) Dockerfiles in the git repository and some instructions how to build them.

> On the other hand, if you set up something that creates binaries automatically on commit, it would make sense to at least point out to such system for other's benefit.

Yes its possible such build to be setup in the Docker Hub/Cloud, Quay.io or some of the other Docker registries and I could help you do it in your account in case you decide that you want such images.

PR #223, opened 2017-10-18
Do not ignore the test errors in TravisCI

velichkov:


.travis.yml
```diff
@@ -3,13 +3,15 @@ compiler:
...
+  - sudo apt-get install -y gcc-multilib lcov libasan* libclang-3.9-dev
   - gem install coveralls-lcov
 script:
     - autoreconf -iv
     - ./configure --enable-Werror --enable-code-coverage
     - make -j8
-    - make check 2>&1 | tail -10000
-    - make distcheck 2>&1 | tail -10000
+    - make check
+    - make distcheck
```
vlm: Without tail Travis won't save enough of the log to be useful. Try `set -o pipefail` prior to make to enable error propagation through the bash pipes.

configure.ac
```diff
@@ -121,6 +121,8 @@ AC_ARG_ENABLE(test-Werror,
   AS_IF([test x$enable_test_werror != xno], [
       TESTSUITE_CFLAGS="${TESTSUITE_CFLAGS} -Werror -W -Wpointer-arith"
   ])
+AX_CHECK_COMPILE_FLAG([-Wno-format],
+    [SKELETONS_CFLAGS="$SKELETONS_CFLAGS -Wno-format"])
```
vlm: The formats should work without warnings. What kind of warnings do you see that require this flag? Do you refer to `%zd/%zu` in -std=c89 mode?

.travis.yml
velichkov: Yes, I already did but with tail the build terminates due to timeout

https://travis-ci.org/velichkov/asn1c/jobs/289640720#L1679-L1686

actually `make check` and `make distcheck` are redundant as `distcheck` always invoke `check` internally

skeletons/NativeReal.c
```diff
@@ -361,15 +361,24 @@ NativeReal_decode_oer(const asn_codec_ctx_t *opt_codec_ctx,
...
-                ASN__DECODE_FAILED;
-            break;
-        case sizeof(float):
-            if(NativeReal__set(td, sptr, *(const float *)scratch) < 0)
-                ASN__DECODE_FAILED;
-            break;
+            case sizeof(double):
+                {
+                    double tmp;
+                    memcpy(&tmp, scratch, sizeof(double));
```
vlm: That's one way to do it. Alternatively, the scratch could be a union of a buffer, double, float.

Do you have a particular reason to prefer this form?

configure.ac
velichkov: Yes, exactly. Here is the travis log
https://travis-ci.org/velichkov/asn1c/jobs/289572223#L1129-L1137

skeletons/OCTET_STRING.c
```diff
@@ -1748,7 +1748,7 @@ OCTET_STRING_free(const asn_TYPE_descriptor_t *td, void *sptr,
     case ASFM_FREE_UNDERLYING_AND_RESET:
         memset(sptr, 0,
                td->specifics
-                   ? ((const asn_OCTET_STRING_specifics_t *)(td->specifics))
+                   ? (size_t)((const asn_OCTET_STRING_specifics_t *)(td->specifics))
```
vlm: This signedness mismatch better handled by a patch in a different place, here: https://github.com/vlm/asn1c/commit/19e12b5597e490bc9113f4fca2be9e2ea75dcdd5

velichkov: Ok, I'm going to drop this commit and force push

skeletons/oer_support.c
```diff
@@ -35,7 +35,7 @@ oer_fetch_length(const void *bufptr, size_t size, size_t *len_r) {
     }
 
     len_len = (first_byte & 0x7f);
-    if((1 + len_len) > size) {
+    if((size_t)(1 + len_len) > size) {
```
vlm: Why this is necessary? len_len is already unsigned.

velichkov: Here is the travis log https://travis-ci.org/velichkov/asn1c/jobs/289628105#L1138-L1140
>   CC       oer_support.lo
oer_support.c: In function ‘oer_fetch_length’:
oer_support.c:38:22: error: comparison between signed and unsigned integer expressions [-Werror=s

.travis.yml
vlm: Got it, thank you.

Consider removing make check, then.

configure.ac
vlm: OK.

skeletons/oer_support.c
vlm: Oh. You could just switch uint8_t to size_t as len_len's type, then. Avoids confusion.

skeletons/NativeReal.c
velichkov: I read various stack overflow posts  :blush: and decided to play it safe
https://stackoverflow.com/questions/7630150/dereferencing-type-punned-pointer-will-break-strict-aliasing-rules#comment9265588_7630162
https://stackoverflow.com/questions/44245618/how-to-legally-use-type-punning-with-unions-to-cast-between-variations-of-struct?rq=1

I could try to rewrite it with a union if you prefer it this way

vlm: No, as long as you picked it according to some reason, I am good with it.

vlm:
Would you please also consider adding `--enable-test-Werror` to the `./configure` flags.

vlm:
Overall, looks good, I'll merge after you say it's done.

tests/tests-skeletons/check-GeneralizedTime.c
```diff
@@ -190,8 +190,8 @@ check_fractions() {
 
 static void
 compare(int lineno, int cmp_control, const char *astr, const char *bstr) {
-    GeneralizedTime_t a = {(uint8_t *)strdup(astr), strlen(astr)};
-    GeneralizedTime_t b = {(uint8_t *)strdup(bstr), strlen(bstr)};
+    GeneralizedTime_t a = {(uint8_t *)strdup(astr), strlen(astr), {0}};
+    GeneralizedTime_t b = {(uint8_t *)strdup(bstr), strlen(bstr), {0}};
```
vlm: Why is this? Some other warning? They're supposed to be 0-initialized on their own.

tests/tests-skeletons/check-INTEGER.c
```diff
@@ -133,7 +133,7 @@ check_unsigned(uint8_t *buf, int size, unsigned long check_long, int check_ret)
 	assert(shared_scratch_start < scratch + sizeof(scratch));
 	assert(ret == 0);
 	ret = snprintf(verify, sizeof(verify), "%lu", check_long);
-	assert(ret < sizeof(verify));
+	assert(ret < (int)sizeof(verify));
```
vlm: `assert(ret >= 0 && (size_t)ret < sizeof(verify));`

Otherwise (int) does narrowing, and not checking <0 doesn't check for system errors.

tests/tests-skeletons/check-OIDs.c
```diff
@@ -92,13 +92,13 @@ check_ROID(int lineno, uint8_t *buf, size_t len, unsigned *ck_buf,
...
     assert(alen > 0);
-	assert(alen == ck_len);
+	assert((size_t)alen == ck_len);
 
 	/*
 	 * Make sure they are equivalent.
 	 */
 	printf("RELATIVE_OID_get_arcs() => {");
-	for(i = 0; i < alen; i++) {
+	for(i = 0; i < (size_t)alen; i++) {
```
vlm: Please make sure alen is `ssize_t` in the first place.

tests/tests-skeletons/check-GeneralizedTime.c
velichkov: Because of the clang https://travis-ci.org/vlm/asn1c/jobs/289782301#L1179-L1186
I'm done for today, I will continue tomorrow

PR #224, opened 2017-10-19
Fix s1ap code generation

brchiu:
This series of pull requests is to add the capability of processing S1AP's ASN.1 and generate C code.
Partially solve #185. 

Due to APER is not supported in current asn1c yet, verification of S1AP codec's correctness still be pending.

PR #227, opened 2017-10-23
Fix duplicate 'asn_VAL_23_ignore' generated from RANAP ASN.1

brchiu:
Information object set RANAP-ELEMENTARY-PROCEDURES is a union of
RANAP-ELEMENTARY-PROCEDURES-CLASS-1 and RANAP-ELEMENTARY-PROCEDURES-CLASS-2.

Each time _asn1f_assign_cell_value() parses an object, value of an
incremented counter is assigned to '_type_unique_index' field for
distinquishing different items.

However, these two information object sets have their own counters.
It has chance that generated variables have the same name and also the same
'_type_unique_index'. And they collide if they join up into one set.

S1AP's ASN.1 is lucky without this problem, while RANAP has.

By adding the number of items of the first set to '_type_unique_index' field
of items of the second set can solve this issue.


vlm:
Please add a test.

vlm:
I would rather have a two smaller tests each verifying a particular thing, than one huge test verifying several things at once.

brchiu:
Both are test union of ios. No difference.

If I found there is name conflict of generated variables for RANAP ASN.1 before I submited #224, this pull request will be merged into #224 instead of a separate one.

vlm:
Merged, thank you!

PR #232, opened 2017-10-29
Correctly report the required output buffer size for PER encoding

### LINKED ISSUE #231: Basic UPER encoding fails to determine required buffer length
MacOS 10.12.6, Xcode-9.0.1. 

ASN.1 file:
```
Test2b DEFINITIONS ::=
BEGIN

Test2b ::= SEQUENCE {

      count INTEGER (40..20000),
      data  OCTET STRING (SIZE (40..20000))
}

END
```

Source code where it fails:
```
   printf("\nAttempt to produce and consume Basic UPER encoding...\n\n");

    /*
     * Now let's try Basic UPER encoding that's supposed to produce
     * the most compact encoding
     */

    /* Experiment with trying to ask how big  buffer we need */
    /* attempt to make asn1c code tell us how big the buffer should be */
    er = uper_encode(&asn_DEF_Test2b, NULL, t_p, NULL, 0);
    if (er.encoded < 0) {
        fprintf(stderr,
                "It refused to tell buffer size needed for UPER-encoding (rc=%ld)\n",
                er.encoded);
        goto end;
    } else {
        buf_len = ((er.encoded + 7)/8); /* convert length in bits to bytes */
        printf("UPER-encoding Test2b_t PDU requires %ld bytes (%ld bits)\n",
               buf_len, er.encoded);
    }
    per_buf = calloc (1, buf_len); /* allocating the determined buffer size... */
    er = uper_encode_to_buffer(&asn_DEF_Test2b, NULL, t_p, per_buf, buf_len);
    if (er.encoded < 0) {
        fprintf(stderr,
                "Failed to UPER-encode \"Basic\" Test2b_t PDU... (rc=%ld)\n",
                er.encoded);
        goto end;
    }
    printf("Successfully UPER-encoded \"Basic\" Test2b_t PDU (%ld bits packed in %ld bytes)\n",
           er.encoded,  ((er.encoded+7)/8));
```

Screen output:
```
. . . . .
Attempt to produce and consume Basic UPER encoding...

Encoding Test2b as SEQUENCE (UPER) (constr_SEQUENCE.c:1372)
first_extension = -1, elements = 2 (constr_SEQUENCE.c:1426)
About to encode INTEGER (constr_SEQUENCE.c:1435)
Encoding Test2b->count:INTEGER (constr_SEQUENCE.c:1458)
Encoding NativeInteger INTEGER 43 (UPER) (NativeInteger.c:306)
Value 43 (2b/1) lb 40 ub 20000 fix (INTEGER.c:754)
Encoding integer 43 (3) with range 15 bits (INTEGER.c:771)
[PER put 15 bits 3 to 0x7fff5fbfeca8+0 bits] (asn_bit_data.c:200)
[PER out 15 3/3 (t=0,o=15) 18&0=0] (asn_bit_data.c:243)
[PER out 6/6 => 00 buf+0] (asn_bit_data.c:274)
Freeing INTEGER as a primitive type (asn_codecs_prim.c:125)
About to encode OCTET STRING (constr_SEQUENCE.c:1435)
Encoding Test2b->data:OCTET STRING (constr_SEQUENCE.c:1458)
Encoding OCTET STRING into 43 units of 8 bits (40..20000, effective 15) (OCTET_STRING.c:1581)
Encoding 43 bytes (3), length in 15 bits (OCTET_STRING.c:1608)
[PER put 15 bits 3 to 0x7fff5fbfeca8+15 bits] (asn_bit_data.c:200)
[PER out 15 3/3 (t=7,o=22) 6&fe=6] (asn_bit_data.c:243)
[PER out 12/c => 06 buf+1] (asn_bit_data.c:274)
Squeezing 43 characters into (0..0):8 (1 bpc) (OCTET_STRING.c:1285)
[PER put 24 bits b27f9a to 0x7fff5fbfeca9+22 bits] (asn_bit_data.c:200)
[PER out 24 11698074/b27f9a (t=6,o=30) c&fc=c] (asn_bit_data.c:243)
[PER out 46792296/2c9fe68 => 0e buf+3] (asn_bit_data.c:274)
[PER put 24 bits ebecc6 to 0x7fff5fbfecab+30 bits] (asn_bit_data.c:200)
[PER out 24 15461574/ebecc6 (t=6,o=30) 68&fc=68] (asn_bit_data.c:243)
[PER out 61846296/3afb318 => 6b buf+6] (asn_bit_data.c:274)
[PER put 24 bits 1b5f76 to 0x7fff5fbfecae+30 bits] (asn_bit_data.c:200)
[PER out 24 1793910/1b5f76 (t=6,o=30) 18&fc=18] (asn_bit_data.c:243)
[PER out 7175640/6d7dd8 => 18 buf+9] (asn_bit_data.c:274)
[PER put 24 bits bfd70b to 0x7fff5fbfecb1+30 bits] (asn_bit_data.c:200)
[PER out 24 12572427/bfd70b (t=6,o=30) d8&fc=d8] (asn_bit_data.c:243)
[PER out 50289708/2ff5c2c => da buf+12] (asn_bit_data.c:274)
[PER put 24 bits a3cf80 to 0x7fff5fbfecb4+30 bits] (asn_bit_data.c:200)
[PER out 24 10735488/a3cf80 (t=6,o=30) 2c&fc=2c] (asn_bit_data.c:243)
[PER out 42941952/28f3e00 => 2e buf+15] (asn_bit_data.c:274)
[PER put 24 bits 1ce953 to 0x7fff5fbfecb7+30 bits] (asn_bit_data.c:200)
[PER out 24 1894739/1ce953 (t=6,o=30) 0&fc=0] (asn_bit_data.c:243)
[PER out 7578956/73a54c => 00 buf+18] (asn_bit_data.c:274)
[PER put 24 bits 4b41be to 0x7fff5fbfecba+30 bits] (asn_bit_data.c:200)
[PER out 24 4932030/4b41be (t=6,o=30) 4c&fc=4c] (asn_bit_data.c:243)
[PER out 19728120/12d06f8 => 4d buf+21] (asn_bit_data.c:274)
[PER put 24 bits 106327 to 0x7fff5fbfecbd+30 bits] (asn_bit_data.c:200)
[PER out 24 1073959/106327 (t=6,o=30) f8&fc=f8] (asn_bit_data.c:243)
[PER out 4295836/418c9c => f8 buf+24] (asn_bit_data.c:274)
[PER put 24 bits 901641 to 0x7fff5fbfecc0+30 bits] (asn_bit_data.c:200)
[PER out 24 9442881/901641 (t=6,o=30) 9c&fc=9c] (asn_bit_data.c:243)
[PER out 37771524/2405904 => 9e buf+27] (asn_bit_data.c:274)
[PER put 24 bits cc1652 to 0x7fff5fbfecc3+30 bits] (asn_bit_data.c:200)
[PER output 30 complete + 0] (asn_bit_data.c:219)
Process 98307 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
    frame #0: 0x0000000000000000
error: memory read failed for 0x0
Target 0: (m1b) stopped.
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
  * frame #0: 0x0000000000000000
    frame #1: 0x0000000100001e1e m1b`asn_put_few_bits(po=0x00007fff5fbfec90, bits=13375058, obits=24) at asn_bit_data.c:220 [opt]
    frame #2: 0x00000001000020ff m1b`asn_put_many_bits(po=0x00007fff5fbfec90, src="?y\b#d\x88??'\x95R?, nbits=104) at asn_bit_data.c:293 [opt]
    frame #3: 0x0000000100005389 m1b`OCTET_STRING_encode_uper(td=0x0000000100022598, constraints=<unavailable>, sptr=<unavailable>, po=0x00007fff5fbfec90) at OCTET_STRING.c:1612 [opt]
    frame #4: 0x000000010001d6f4 m1b`SEQUENCE_encode_uper(td=<unavailable>, constraints=<unavailable>, sptr=0x0000000100102840, po=0x00007fff5fbfec90) at constr_SEQUENCE.c:1459 [opt]
    frame #5: 0x000000010000bf24 m1b`uper_encode(td=0x00000001000229c0, constraints=<unavailable>, sptr=0x0000000100102840, cb=<unavailable>, app_key=<unavailable>) at per_encoder.c:27 [opt]
    frame #6: 0x0000000100001054 m1b`try_string(input_str=<unavailable>, len=<unavailable>) at m1b.c:298 [opt]
    frame #7: 0x0000000100001370 m1b`main(argc=<unavailable>, argv=<unavailable>) at m1b.c:392 [opt]
    frame #8: 0x00007fffa881d235 libdyld.dylib`start + 1
    frame #9: 0x00007fffa881d235 libdyld.dylib`start + 1
(lldb)
```

**Proposed solution:**
~~https://github.com/mouse07410/asn1c/commit/53201831731d98d56d28524177ebb32df9710845~~
and you'll probably need something similar for OER.
This problem showed up in my fork last April, and @velichkov fixed it nicely then.

### Update
**This** is the solution I've successfully tested with the current master:
```
diff --git a/skeletons/asn_bit_data.c b/skeletons/asn_bit_data.c
index fe4b89ba..de651e95 100644
--- a/skeletons/asn_bit_data.c
+++ b/skeletons/asn_bit_data.c
@@ -217,7 +217,7 @@ asn_put_few_bits(asn_bit_outp_t *po, uint32_t bits, int obits) {
                complete_bytes = (po->buffer - po->tmpspace);
                ASN_DEBUG("[PER output %ld complete + %ld]",
                        (long)complete_bytes, (long)po->flushed_bytes);
-               if(po->output(po->tmpspace, complete_bytes, po->op_key) < 0)
+               if(po->output && po->output(po->tmpspace, complete_bytes, po->op_key) < 0)
                        return -1;
                if(po->nboff)
                        po->tmpspace[0] = po->buffer[0];
diff --git a/skeletons/per_encoder.c b/skeletons/per_encoder.c
index 17e5fce5..bfe230f2 100644
--- a/skeletons/per_encoder.c
+++ b/skeletons/per_encoder.c
@@ -152,6 +152,9 @@ _uper_encode_flush_outp(asn_per_outp_t *po) {
                buf++;
        }
 
-       return po->output(po->tmpspace, buf - po->tmpspace, po->op_key);
+       if (po->output) {
+               return po->output(po->tmpspace, buf - po->tmpspace, po->op_key);
+       }
+       return 0;
 }
 
```
@vlm I'd appreciate if you could ~~apply this~~ merge #232 to fix this - being unable to request required output buffer length is very inconvenient.

mouse07410:
This avoids application code crashing with SEGV upon attempt to learn the required size of the output buffer for PER encoding.

Resolves #231

### DIFF
--- a/skeletons/asn_bit_data.c
+++ b/skeletons/asn_bit_data.c
@@ -217,7 +217,7 @@ asn_put_few_bits(asn_bit_outp_t *po, uint32_t bits, int obits) {
 		complete_bytes = (po->buffer - po->tmpspace);
 		ASN_DEBUG("[PER output %ld complete + %ld]",
 			(long)complete_bytes, (long)po->flushed_bytes);
-		if(po->output(po->tmpspace, complete_bytes, po->op_key) < 0)
+		if(po->output && po->output(po->tmpspace, complete_bytes, po->op_key) < 0)
 			return -1;
 		if(po->nboff)
 			po->tmpspace[0] = po->buffer[0];
--- a/skeletons/per_encoder.c
+++ b/skeletons/per_encoder.c
@@ -152,6 +152,9 @@ _uper_encode_flush_outp(asn_per_outp_t *po) {
 		buf++;
 	}
 
-	return po->output(po->tmpspace, buf - po->tmpspace, po->op_key);
+        if (po->output) {
+                return po->output(po->tmpspace, buf - po->tmpspace, po->op_key);
+        }
+        return 0;
 }
 


vlm:
1. Please use the `asn_encode()` from `asn_application.h`, which returns the amount of bytes required.

2. This patch is suspicious to me because it silently loses data. While it helps you in your case (however see `asn_encode()`), it introduces a possibility of losing data somewhere else if `po->output` is not set up properly. I'd prefer it to fail early in this case.

What do you think?

vlm:
1. Pardon my confusion, would you point out where it contradicts the existing documentation?

2. Regarding the successful encoding, how do you check the result of snprintf() as to whether it printed the stuff correctly or not? `asn_encode_to_buffer()` has the same semantics.

mouse07410:
> Pardon my confusion, would you point out where it contradicts the existing documentation?

Much to my shame, I did not keep the old version of the `asn1c-usage.pdf`, and in the current one I cannot find the appropriate reference. But I don't think I just invented it that invoking `xxx_encode()` with NULL buffer should return the required buffer size. So I must've read it somewhere in the asn1c docs.  :-(

> Regarding the successful encoding, how do you check the result of `snprintf()` as to whether it printed the stuff correctly or not?

Interestingly, in all my years I have never used `snprintf()` so far. Mainly for security reasons (risk of stepping on somebody else's memory is non-zero :).

craig65535:
@mouse07410 snprintf is perfectly safe if you specify the output buffer size correctly. It never writes more than that number of bytes, it always null-terminates the string, and the return value lets the caller know if the output was truncated. As long as you're not doing something like using user input as a format string, I think there are no security reasons to avoid it.

mouse07410:
> snprintf is perfectly safe if you specify the output buffer size correctly

Thanks, good to know. 

Still, so far I managed to survive without using it. ;-). It is tangential to our issue anyway. ;-)

vlm:
1. Providing NULL instead of the collback was introduced for DER, since it recursively needs to know the size of underlying structures before encoding the length determinants. The problem with that approach is that it introduces two code paths: one is to determine the size and another is to encode. This is error-prone and nasty, I don't like it. However, it allows to encode a little bit faster, since instead of sorting in some places to determine its size we just return its size without sorting anything. I could have removed this usage pattern (using NULL as a value of `callback` function), but it would significantly slow down DER encoding, which is not that fast anyway.

2. I decided not introduce that nasty thing to PER. Instead, if you provide a function which captures the length instead of saving bytes, it will return what you seek. This is exactly what `asn_encode_to_buffer()` does: it invokes encoder with a custom callback which just counts bytes. This approach doesn't slow down encoding, unlike in the DER example above.

3. The `asn_encode_to_buffer()` is modeled after well-known `snprintf()`, which is, as @craig65535 says, quite memory-safe. In order to figure out that you've got everything, just make sure that the return value doesn't exceed the size of your output buffer. I'll introduce another function, `asn_encode_to_new_buffer()` to allow not thinking about such things and just get a newly allocated buffer of a correct size.

4. As I said, if `po->output` value is ignored, it will lose data in the application code. This wouldn't be asn1c's code, this would be applications author's code. Since PER encodings are usually so compact (they encode into less than 32 bytes for many real world uses), there's a very high chance that application writers would not initialize callback properly and it would still work for them for quite some time. Until it fails on some long-ish object. But it'll fail silently and/or with a memory corruption, not with a clear crash. Therefore I am reluctant to introduce this mode.

mouse07410:
I hear you. In the process of testing the new master.

Out of curiosity: is there a way to capture the required size in *exact* bits (besides capturing the required size in whole bytes)?

PR #226, opened 2017-10-22
Merge APER implementation from mouse07410's repository

brchiu:
Merge APER implementation from @mouse07410's repository.

Due to there is no APER encoder and decoder for OPEN TYPE, so `s1ap-dump` built is still not able to decode S1AP messages.

PR #234, opened 2017-11-02
Fix s1ap code generation (2nd pull reuqest)

brchiu:
Fix most S1AP decoding issue till now. See discussion in #185.

After merged with #226, some S1AP sample message can be decoded.


mouse07410:
This PR has a (small) conflict with #234 in file tests/tests-asn1c-compiler/50-constraint-OK.asn1.-Pgen-PER.

brchiu:
@mouse07410, you can refer to my local branch https://github.com/brchiu/asn1c/tree/s1ap to see how to resolve it.

This conflict comes from my commits : https://github.com/brchiu/asn1c/commit/e436f61aa7fcd336e65c317b425b098713d67192 and https://github.com/brchiu/asn1c/commit/8d0f6857eb032b04558a12d9a6757efd717e8f46 which generate some constant definition not available in current implementation.

I think my tweak is quite ugly and perhaps requires to be rewritten in a better way.

Issue #217, opened 2017-10-10
Failed decoding extensions in RRC


zaheermot:
There is an issue decoding RRC messages, specifially with extension groups.

bytes: 08 21 be 48 16 01 00 03  42 2a c1
./rrc-dump -dd -1 -p UL-DCCH-Message <msg.per (hextob of above)
asn1: http://code.metager.de/source/xref/wireshark/asn1/lte-rrc/EUTRA-RRC-Definitions.asn

MeasResults ::=           SEQUENCE {
  measId                MeasId,
  measResultPCell         SEQUENCE {
    rsrpResult              RSRP-Range,
    rsrqResult              RSRQ-Range
  },
  measResultNeighCells        CHOICE {
    measResultListEUTRA         MeasResultListEUTRA,
    measResultListUTRA          MeasResultListUTRA,
    measResultListGERAN         MeasResultListGERAN,
    measResultsCDMA2000         MeasResultsCDMA2000,
    ...
  }                                   OPTIONAL,
  ...,
  [[  measResultForECID-r9        MeasResultForECID-r9      OPTIONAL
  ]],
  [[  locationInfo-r10          LocationInfo-r10        OPTIONAL,
    measResultServFreqList-r10      MeasResultServFreqList-r10    OPTIONAL
  ]]
}

l=3 (per_support.c:60)
Extensions 3 present in MeasResults (constr_SEQUENCE.c:1200)
  [PER got  3<=51 bits => span 40 +4[8..56]:16 (48) => 0x6] (asn_bit_data.c:114)
Read in extensions bitmap for MeasResults of 3 bits (c0..) (constr_SEQUENCE.c:1215)
  [PER got  1<= 3 bits => span 1 +0[1..3]:c0 (2) => 0x1] (asn_bit_data.c:114)
Decoding member measResultForECID-r9 in MeasResults (nil) (constr_SEQUENCE.c:1243)
Getting open type MeasResultForECID-r9... (per_opentype.c:78)
  [PER got  8<=48 bits => span 48 +5[8..48]:01 (40) => 0x1] (asn_bit_data.c:114)
  [PER got  8<=40 bits => span 56 +6[8..40]:00 (32) => 0x0] (asn_bit_data.c:114)
Getting open type MeasResultForECID-r9 encoded in 1 bytes (per_opentype.c:104)
Decoding MeasResultForECID-r9 as SEQUENCE (UPER) (constr_SEQUENCE.c:1100)
[...]

Issue #207, opened 2017-09-19
Failure to decode extension Addition Group [[ .. ]] for ASN.1 PER bit aligned. 


dbressan2:
Hi 

I am testing 0.9.28 
I can generate LTE RRC decoder however all messages including extension Addition Group [[ ]]] fail decoding. 
Error looks like 
AD: Processing RRC_conn_setup.txt
AD: Decoding 39 bytes
AD: decode(0) consumed 0+0b (39), code 2
AD: Clean up partially decoded DL-CCCH-Message
AD: ofp 1, no=0, oo=0, dbl=0
RRC_conn_setup.txt: Decode failed past byte 0: Input processing error

Now if I replace the [[ ]] construct with equivalent ASN.1 syntax extensionAdditionGroup SEQUENCE {}, decoding is successful. 

Issue is not blocking but annoying as I cannot directly work from 3GPP ASN.1 syntax. 

Example: original 3GPP ASN.1 (decoding fails):
[[	-- DL configuration as well as configuration applicable for DL and UL
		csi-RS-ConfigNZPToReleaseList-r11
									CSI-RS-ConfigNZPToReleaseList-r11	OPTIONAL,		-- Need ON
		csi-RS-ConfigNZPToAddModList-r11
									CSI-RS-ConfigNZPToAddModList-r11	OPTIONAL,		-- Need ON
		csi-RS-ConfigZPToReleaseList-r11	
									CSI-RS-ConfigZPToReleaseList-r11	OPTIONAL,		-- Need ON
		csi-RS-ConfigZPToAddModList-r11	CSI-RS-ConfigZPToAddModList-r11	OPTIONAL,		-- Need ON
		epdcch-Config-r11				EPDCCH-Config-r11				OPTIONAL,		-- Need ON
		pdsch-ConfigDedicated-v1130		PDSCH-ConfigDedicated-v1130		OPTIONAL,		-- Need ON
	-- UL configuration
		cqi-ReportConfig-v1130			CQI-ReportConfig-v1130			OPTIONAL,		-- Need ON
		pucch-ConfigDedicated-v1130		PUCCH-ConfigDedicated-v1130		OPTIONAL,		-- Need ON
		pusch-ConfigDedicated-v1130		PUSCH-ConfigDedicated-v1130		OPTIONAL,		-- Need ON
		uplinkPowerControlDedicated-v1130
									UplinkPowerControlDedicated-v1130	OPTIONAL		-- Need ON
	]],

[...]

Issue #198, opened 2017-09-05
Bug on extension of sequence or choice


lwmqwer:
Dear Mr. Walin,
I am trying to use the asn1c to decode some 3GPP messages.
But I found some promblem in the extention decoing in sequence or choise in per decoder.
For instance, 
Example ::= sequence {
num INTERGER,
...,
[[ num2 INTERGER OPTIONAL,
]]
}
the num2 is treat as opentype, but it need extra 1 bit check in the opentype for the optional.

In this exampe, the num2 and num3 should deal as one opentype but shoud be decode as  a sequence {num2, num3}
Example2 ::= sequence {
num INTERGER,
...,
[[ num2 INTERGER OPTIONAL,
   num3 INTERGER OPTIONAL
]]
}
I have check X.691 about this.

brchiu:
@lwmqwer, I submitted a pull request #238 to address this extension group `[[...]]` issue. FYI.

PR #238, opened 2017-11-05
Tentative solution to extension group (only for PER)

brchiu:
With the trick to use `SEQUENCE` type to hold members of a extension group, the generated `lte-rrc-dump` can decode `bytes: 08 21 be 48 16 01 00 03 42 2a c1` mentioned in #217.

```
$ ./lte-rrc-dump -iper -p UL-DCCH-Message ./sample-UL-DCCH-Message-1.per 
<UL-DCCH-Message>
    <message>
        <c1>
            <measurementReport>
                <criticalExtensions>
                    <c1>
                        <measurementReport-r8>
                            <measResults>
                                <measId>4</measId>
                                <measResultPCell>
                                    <rsrpResult>62</rsrpResult>
                                    <rsrqResult>18</rsrqResult>
                                </measResultPCell>
                                <(null)>
                                </(null)>
                                <(null)>
                                    <measResultServFreqList-r10>
                                        <MeasResultServFreq-r10>
                                            <servFreqId-r10>1</servFreqId-r10>
                                            <measResultSCell-r10>
                                                <rsrpResultSCell-r10>43</rsrpResultSCell-r10>
                                                <rsrqResultSCell-r10>1</rsrqResultSCell-r10>
                                            </measResultSCell-r10>
                                        </MeasResultServFreq-r10>
                                    </measResultServFreqList-r10>
                                </(null)>
                            </measResults>
                        </measurementReport-r8>
                    </c1>
                </criticalExtensions>
            </measurementReport>
        </c1>
    </message>
</UL-DCCH-Message>
```
Though @vlm states in [his comment](https://github.com/vlm/asn1c/issues/217#issuecomment-335714513), that this trick only applies to PER encoding. 
Since it is ASN.1 of LTE-RRC that current users concern about support of extension group, based on #198, #207 and #217. It might be temporarily sufficient for our need.

Improvement and/or even re-implementation of this feature is welcomed.


libasn1compiler/asn1c_misc.c
```diff
@@ -61,8 +61,13 @@ asn1c_make_identifier(enum ami_flags_e flags, asn1p_expr_t *expr, ...) {
 		/*
 		 * Estimate the necessary storage size
 		 */
-		if(expr->Identifier == NULL)
-			return "Member";
+		if(expr->Identifier == NULL) {
+			if(storage_size < 10) {
```
craig65535: 10 is not sufficient to store `MemberNNNNNNNNNNN\0`. I think you mean 20 here (which is what you used below).

libasn1compiler/asn1c_misc.c
```diff
@@ -61,8 +61,13 @@ asn1c_make_identifier(enum ami_flags_e flags, asn1p_expr_t *expr, ...) {
...
+				storage = calloc(1, 20);  /* reserve space for 'Memberxxx' */
+			}
+			sprintf(storage, "Member%d", expr->_type_unique_index);
```
craig65535: Only do this if `storage != NULL`. Also I would suggest `snprintf(storage, storage_size, ...` for safety.

libasn1compiler/asn1c_misc.c
```diff
@@ -61,8 +61,13 @@ asn1c_make_identifier(enum ami_flags_e flags, asn1p_expr_t *expr, ...) {
 		/*
 		 * Estimate the necessary storage size
 		 */
-		if(expr->Identifier == NULL)
-			return "Member";
+		if(expr->Identifier == NULL) {
+			if(storage_size < 10) {
+				storage = calloc(1, 20);  /* reserve space for 'Memberxxx' */
```
craig65535: This is a memory leak. Free `storage` first. You need to update `storage_size` as well as it is also static.

libasn1compiler/asn1c_misc.c
brchiu: Thank you for reviewing !

It was almost 2:00 am at my place so I don't know what I was thinking then. 😢

libasn1compiler/asn1c_misc.c
brchiu: I meant to use realloc, but .......

Thanks for the advice to replace it with snprintf()

libasn1compiler/asn1c_misc.c
brchiu: Thanks, I meant to use realloc, but....

I will fix it.

brchiu:
For aforementioned message decoded, I am not sure where the following decoding result comes from.
It might worthy further check.
```
        <(null)>
        </(null)>
```

vlm:
@brchiu this result is something that I warned you about in the first place: it applies to PER. What you have is XER, so something else is needed, such as introduction of anonymous substructures concept. That would entail changing the ber and xer encoders for SEQUENCE.

libasn1compiler/asn1c_misc.c
```diff
@@ -61,8 +61,18 @@ asn1c_make_identifier(enum ami_flags_e flags, asn1p_expr_t *expr, ...) {
 		/*
 		 * Estimate the necessary storage size
 		 */
-		if(expr->Identifier == NULL)
-			return "Member";
+		if(expr->Identifier == NULL) {
+			if(storage_size < 20) {
+				storage = realloc(storage, 20);  /* reserve space for 'Memberxxx' */
```
craig65535: Since you're not reusing the contents of the buffer, I would recommend free+malloc here instead. It also avoids the memory leak that would happen if realloc returned NULL.

brchiu:
@vlm, thanks for explanation.

I understand that there are corresponding modification required for encoding rules other than PER. I sent this premature pull request is for those who only care about 3GPP protocols that use PER.  I am also one of them. We don't care much whether this will break BER or XER...., at least if they are broken with a known way.

By the way, this also bring up another topic that if we can remove unused codec and reduce code size.
This might be necessary for memory limited devices.
Long time ago I tried to add conditional compilation for removing unused data structure, but it need to be re-implemented since many things changed.
Could you advice that how I can proceed ?

vlm:
Let me do a release in the next couple of days first, then will coordinate.

mouse07410:
When #226 is applied, this PR fails tests. Here's an example :
```
. . . . .
Checking ../../tests/tests-asn1c-compiler/159-extension-groups-OK.asn1 against ../../tests/tests-asn1c-compiler/159-extension-groups-OK.asn1.-Pgen-PER
Generated asn_constant.h
. . . . .
@@ -352,6 +360,8 @@
 xer_type_encoder_f ID_V4_encode_xer;
 per_type_decoder_f ID_V4_decode_uper;
 per_type_encoder_f ID_V4_encode_uper;
+per_type_decoder_f ID_V4_decode_aper;
+per_type_encoder_f ID_V4_encode_aper;
 
 /*** <<< CODE [ID-V4] >>> ***/
 
@@ -436,29 +446,29 @@
         * possible extensions are below.
         */
        struct Member5 {
-               ID_V1_t *id_v1  /* OPTIONAL */;
-               OCTET_STRING_t  *data_v1        /* OPTIONAL */;
+               ID_V1_t *id_v1; /* OPTIONAL */
+               OCTET_STRING_t  *data_v1;       /* OPTIONAL */
                
                /* Context for parsing across buffer boundaries */
                asn_struct_ctx_t _asn_ctx;
        } *Member5;
        struct Member8 {
-               ID_V2_t *id_v2  /* OPTIONAL */;
-               OCTET_STRING_t  *data_v2        /* OPTIONAL */;
+               ID_V2_t *id_v2; /* OPTIONAL */
+               OCTET_STRING_t  *data_v2;       /* OPTIONAL */
. . . . .
```

brchiu:
This conflict comes from https://github.com/brchiu/asn1c/commit/1949e868072a6adc3a709c7a724a6837d00d17db which place the generated semicolon before generated comment of the fields, which makes the code like tranditional convention.

You can either skip this commit or resolve it by regenerating test case result using  './check-parsing.sh regenerate' under 'tests/test-asn1c-compiler' directory.

brchiu:
With newest update commits, which mimics OAI's asnfix.c of generate_asn1 script, option groups are given names as `ext1`, `ext2`, .... etc. 

```
$ ./lte-rrc-dump -iper -per-nopad -p UL-DCCH-Message ./sample-UL-DCCH-Message-1.per 
<UL-DCCH-Message>
    <message>
        <c1>
            <measurementReport>
                <criticalExtensions>
                    <c1>
                        <measurementReport-r8>
                            <measResults>
                                <measId>4</measId>
                                <measResultPCell>
                                    <rsrpResult>62</rsrpResult>
                                    <rsrqResult>18</rsrqResult>
                                </measResultPCell>
                                <ext1>
                                </ext1>
                                <ext2>
                                    <measResultServFreqList-r10>
                                        <MeasResultServFreq-r10>
                                            <servFreqId-r10>1</servFreqId-r10>
                                            <measResultSCell-r10>
                                                <rsrpResultSCell-r10>43</rsrpResultSCell-r10>
                                                <rsrqResultSCell-r10>1</rsrqResultSCell-r10>
                                            </measResultSCell-r10>
                                        </MeasResultServFreq-r10>
                                    </measResultServFreqList-r10>
                                </ext2>
                            </measResults>
                        </measurementReport-r8>
                    </c1>
                </criticalExtensions>
            </measurementReport>
        </c1>
    </message>
</UL-DCCH-Message>
```

brchiu:
hi, @vlm, I will withdraw this pull request and re-submit one based on several enhancements at branch [velichkov_s1ap_plus_option_group](https://github.com/brchiu/asn1c/commits/velichkov_s1ap_plus_option_group) of my repository, e.g. https://github.com/brchiu/asn1c/commit/f12568d617dbf48497588f8e227d70388fa217c9, https://github.com/brchiu/asn1c/commit/3d76a0399b065c722db814082f0ddd4dae060a22 and https://github.com/brchiu/asn1c/commit/20879ba25205e228642c9ab877065d489e4246bd ... related to implementation of extension group.

PR #241, opened 2017-11-09
Fix 'make check' error under sample.source.RRC

brchiu:
As title.


brchiu:
Modification of `converter-sample.c` exists in #226.
Main purpose of this pull request is for changing `asn_application.c`.
.

PR #243, opened 2017-11-13
Generate Makefile.am.libasn1codec compatible with non-recursive Makef…

ringlej:
…ile.am

The fact that the existing code creates files with names like:
  Makefile.am.libasncodec
  Makefile.am.example

leads me to believe that there is an attempt to create automake compatible
Makefile.am files. However, there are some things that are wrong with the
example generated files. For example, ASN_MODULES_SOURCES variable can't
be used because all *_SOURCES variables are expected to be either a
bin_PROGRAM or lib_LTLIBRARIES, and of course, there is neither:
  bin_PROGRAMS += ASN_MODULES
or
  lib_LTLIBRARIES += ASN_MODULES

This patch does the following:
1) correct the invalid variable names for use with autotools
2) cleanly separate the library generation from an example as1convert prog
3) Add a new option -D <destdir> that will write the generated files to the
   specified directory, and the generated Makefile.am.* will be compatible
   with a non-recursive Makefile.am schema
4) Add a new option -gen-autotools which will create (if they don't exist) a minimal
   example configure.ac and Makefile.am

This will make it much easier to integrate the generated asn1 source files with a larger project

### DIFF
--- a/asn1c/asn1c.c
+++ b/asn1c/asn1c.c
@@ -63,16 +63,19 @@ main(int ac, char **av) {
     int print_arg__fix_n_print = 0; /* Fix and print */
     int warnings_as_errors = 0;     /* Treat warnings as errors */
     char *skeletons_dir = NULL;     /* Directory with supplementary stuff */
+    char destdir[PATH_MAX];         /* Destination directory for generated files */
     asn1p_t *asn = 0;               /* An ASN.1 parsed tree */
     int ret;                        /* Return value from misc functions */
     int ch;                         /* Command line character */
     int i;                          /* Index in some loops */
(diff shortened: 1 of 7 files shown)


ringlej:
It's clear to me now that this patch won't do in it's current form for the unit tests. I will work on this and send another PR

vlm:
@ringlej sounds good. Prefer to send separate self-contained PRs. Such as, the -gen-autotools ought to go as a separate PR from the earlier items.

ringlej:
Would you be ok with one PR with each item as a separate commit?

vlm:
That would work as well.

Also, please make sure you rebase first. I just added a bunch of changes around fdeps.

PR #245, opened 2017-11-14
Cleanup generated makefiles and make it work better with autotools

ringlej:


.gitignore
```diff
@@ -76,7 +76,9 @@ doc/docsrc/*.xdv
 /examples/sample.source.*/lib*.a
 /examples/sample.source.*/.tmp.*
 /examples/sample.source.*/Makefile.am.libasncodec
+/examples/sample.source.*/libasncodec.am
```
vlm: What's your rationale for changing the makefile names?

asn1c/asn1c.c
```diff
@@ -179,6 +186,11 @@ main(int ac, char **av) {
         case 'S':
             skeletons_dir = optarg;
             break;
+        case 'D':
+            strncat(destdir, optarg, PATH_MAX - 1);
+            if(destdir[strlen(destdir)-1] != '/')
+                strncat(destdir, "/", PATH_MAX - 2);
```
vlm: `(PATH_MAX - 2)` argument to `strncat()` does not make sense here.

tests/tests-asn1c-smoke/check-asn1c-smoke.sh
```diff
@@ -8,7 +8,7 @@ top_builddir=${top_builddir:-../..}
 top_srcdir=${top_srcdir:-../..}
 
 cleanup() {
-    rm -rf *.[acho] Makefile.am.* *.txt *.asn
+    rm -rf *.[acho] *.mk *.am *.txt *.asn
```
vlm: This `rm` will clobber the main `Makefile.am` from which this script is being run. Suggest `lib*am` instead, or something to that effect.

libasn1compiler/asn1c_save.c
```diff
@@ -294,6 +294,58 @@ asn1c__save_example_am_makefile(arg_t *arg, const asn1c_dep_chainset *deps, cons
...
+	}
+
+	fd = asn1c_open_file("", confac, "", 0);
+	if(fd == NULL) {
+		perror(confac);
+		return -1;
+	}
+
+	safe_fprintf(fd,
+	             "AC_INIT([asn1covert],[0.1])\n"
```
vlm: Convert.

.gitignore
ringlej: I changed the Makefile.am.example to example.mk to reflect that it is a *plain* makefile snippet rather than a file that shows the characteristics of a makefile snippet that needs to be transformed via automake. Also, *.mk keeps the syntax highlighting in most editors.
This rationale doesn't hold as much water in regards to the Makefile.am.libasncodec though. I will undo these changes.

asn1c/asn1c.c
ringlej: Ok. I misinterpreted the strncat page. I think this works better:
```c
        case 'D':
            strncat(destdir, optarg, PATH_MAX - 2); /* leave room for possible trailing '/' */
            if(destdir[strlen(destdir)-1] != '/')
                strncat(destdir, "/", PATH_MAX - 1);
```

tests/tests-asn1c-smoke/check-asn1c-smoke.sh
ringlej: I will change to `Makefile.am.* *.mk` since I'll be reverting the Makefile.am.libasncodec name change

libasn1compiler/asn1c_save.c
ringlej: Sorry, that was left over from my previous PR. (I thought `asn1convert` was a better name than `convert-example`). I'll change it to `AC_INIT([convert-example],[0.1])` if you'd prefer that

ringlej: Oh... maybe you were referring to the typo instead :)
asn1covert => asn1convert

.gitignore
vlm: Perhaps the least surprising name would be `converter-example.mk`. This way, there's no question about what does what.

asn1c/asn1c.c
```diff
@@ -179,6 +186,11 @@ main(int ac, char **av) {
         case 'S':
             skeletons_dir = optarg;
             break;
+        case 'D':
+            strncat(destdir, optarg, PATH_MAX - 2); /* leave room for possible trailing '/' */
+            if(destdir[strlen(destdir)-1] != '/')
+                strncat(destdir, "/", PATH_MAX - 1);
```
vlm: The third argument to strncat doesn't do what you want it to do.

ringlej: I'm not sure I follow what I'm missing about strncat arg 3...
The following test program produces the expected output, and makes sure that destdir buffer does not overflow:
```c
#include <stdio.h>

#define MAX_CHARS 10
int main()
{
    const char *optarg = "123456789";
    char destdir[MAX_CHARS];
    destdir[0] = '\0';

    strncat(destdir, optarg, MAX_CHARS - 2); /* leave room for possible trailing '/' */
    if(destdir[strlen(destdir)-1] != '/')
        strncat(destdir, "/", MAX_CHARS - 1);
    
    printf("destdir='%s'\n", destdir);

    return 0;
}
```
Produces this output, that I expect:
```
destdir='12345678/'
```

ringlej: Ok. I see now... That line should be:
```c
strncat(destdir, "/", 1);
```
I will fix it.

vlm: Yes. Though look at it, is pretty much useless this way. `strcat()` could do. `strncat()` is also ok though.

libasn1compiler/asn1c_save.c
vlm: Yes, I was referring to the typo.

.gitignore
```diff
@@ -42,7 +42,8 @@ stamp-h*
 
 # /tests/
 /tests/tests-asn1c-smoke/Makefile.am.libasncodec
-/tests/tests-asn1c-smoke/Makefile.am.sample
+/tests/tests-asn1c-smoke/Makefile.am.asn1convert
```
vlm: Why Makefile.am.asn1convert? It is not used elsewhere.

.gitignore
```diff
@@ -42,7 +42,8 @@ stamp-h*
...
+/tests/tests-asn1c-smoke/example.mk
```
vlm: `converter-example.mk`?

.gitignore
ringlej: The `check-asn1c-smoke.sh` script calls `asn1c` compiler, which will produce `/tests/test-asn1c-smoke/Makefile.am.asn1convert`. Even though, nothing is using it, I think we want git to ignore the file

.gitignore
ringlej: Yes. will fix

asn1c/asn1c.c
ringlej: Yes. I see. I've already left room for the trailing '/', so `strcat()` is sufficient. I will change it.

Issue #254, opened 2018-01-07
Several issues encountered during building an application


certificationoperatingcouncil:
Greetings,

Thanks for adding OER encoding/decoding to your ASN.1 C compiler. I am trying to compile and use the following ASN.1 code: https://github.com/certificationoperatingcouncil/TCI_ASN1/tree/master/TCI%20Interface/ASN1 .

I compiled the ASN.1 code with the following switches: asn1c -no-gen-example -fincludes-quoted -no-gen-PER -fcompound-names -gen-OER , I encountered a few problems, I am trying to summarize below:

1-	The includes-quoted switch was not honored, the header files were included using brackets which caused compile errors, I had to manually edit them;

2-	I tried to compile the generated C files, all the files compiled except for 1, I got an error when compiling the Extensions.c file. I think the correct definition for the RepeateRate should be asn_DEF_IEEE_1609_3_WEE_RepeatRate not asn_DEF_RepeatRate. I see that RepeatRate is defined in 2 different ASN.1 files and that was probably the source of confusion. I edited the file and it compiles now. You can find the generated code here: https://drive.google.com/drive/folders/1plZNimFkQyV4rsibpgM9-6I8GROCmOLz ;

3-	Now to building an application, I was using CodeBlocks with WxWidgets to write a Linux application similar to this Windows application (I am the author of this application): https://github.com/certificationoperatingcouncil/TCI_Tools/tree/master/TCI_Msg_Generator_Portable/V1.14. Using g++ created so many errors, I resorted to building a static library of the generated code and linking with it, that worked;

4-	I encountered A problem with the CodeBlocks, so I switched to creating a Visual Studio 2015 C project to utilize its excellent debugging capabilities. I selected a very simple command to test the application and try to understand how to use the generated C code, this command is RequestSutInfo which includes a Boolean value, here is the code: 

// SUT Control
#include "SutInfo.h"
#include "RequestSutAvailability.h"
#include "RequestSutInfo.h"
#include "Restart.h"
#include "Shutdown.h"
#include "TCIMsg.h"
#include "NativeInteger.h"
#include "Time64.h"
#include "Frame.h"
#include "constr_SEQUENCE.h"
#include "TCISutControl.h"
#include <math.h>

void OnRequestSutInfoBClick();

int main (int argc, const char* argv[])
{
   OnRequestSutInfoBClick();    
}

void OnRequestSutInfoBClick()
{
    TCIMsg_t pdu;
    Frame_t  frame;
[...]

PR #255, opened 2018-01-13
Add module name as prefix for field of an information object set when required

brchiu:
If the type of a field of information object set is imported from other module, then modify corresponding parameter to reflect the fact.

For the case stated in item 2 of #254, there are two `RepeatRate` defined in different modules, i.e. 
`TCICommonTypes.asn` and `IEEE-1609-3-WEE`. So a suitable prefix using module name should be added to distinguish between them and two files, i.e. `TCI-CommonTypes_RepeatRate.c` and  `IEEE-1609-3-WEE_RepeatRate.c`, are generated.

For module `IEEE-1609-3-WSA`, it import `RepeatRate` from module `IEEE-1609-3-WEE`. So 

```
static const asn_ioc_cell_t asn_IOS_SrvAdvMsgHeaderTCI-CommonTypes_RepeatRate.h` and  `IEEE-1609-3-WEE_RepeatRate.c` are generated.ExtTypes_1_rows[] = {
	{ "&extRef", aioc__value, &asn_DEF_RefExt, &asn_VAL_1_c_RepeatRate },
	{ "&ExtValue", aioc__type, &asn_DEF_RepeatRate },
...
	{ "&ExtValue", aioc__type, &asn_DEF_AdvertiserIdentifier }
};
```
should be : 
```
static const asn_ioc_cell_t asn_IOS_SrvAdvMsgHeaderTCI-CommonTypes_RepeatRate.h` and  `IEEE-1609-3-WEE_RepeatRate.c` are generated.ExtTypes_1_rows[] = {
	{ "&extRef", aioc__value, &asn_DEF_RefExt, &asn_VAL_1_c_RepeatRate },
	{ "&ExtValue", aioc__type, &asn_DEF_IEEE_1609_3_WEE_RepeatRate },
...
	{ "&ExtValue", aioc__type, &asn_DEF_AdvertiserIdentifier }
};
```

### DIFF
--- a/libasn1compiler/asn1c_misc.c
+++ b/libasn1compiler/asn1c_misc.c
@@ -217,6 +217,11 @@ asn1c_type_name(arg_t *arg, asn1p_expr_t *expr, enum tnfmt _format) {
 			}
 		}
 
+		if(terminal) {
+			expr->module = terminal->module;
+			expr->_mark = terminal->_mark;
+		}
+
 		if(_format == TNF_CTYPE || _format == TNF_CONSTYPE) {
 			/*
 			 * If the component references the type itself,
--- a/skeletons/OPEN_TYPE.c
+++ b/skeletons/OPEN_TYPE.c
@@ -16,7 +16,12 @@ asn_TYPE_operation_t asn_OP_OPEN_TYPE = {
 	OPEN_TYPE_encode_der,
 	OPEN_TYPE_decode_xer,
 	OPEN_TYPE_encode_xer,
+#ifdef ASN_DISABLE_OER_SUPPORT
 	0, 0,	/* No OER support, use "-gen-OER" to enable */
+#else
+	OPEN_TYPE_decode_oer,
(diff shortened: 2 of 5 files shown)


mittwinter:
Basically, in our project we ran into the same issue as reporter in item no. 2 of issue #254 . However, after applying this pull request to master, we still see two issues along the same lines:
1. Correct prefixes are still not generated consistently: Even if now an ambiguous type itself is generated with the correct prefix, not all instances where the type is used as part of other types use the correct prefix.
1. Superfluous prefixes for some unique, unambiguous types are generated, yet also not used consistently throughout the generated code.

I know that the above description of these issues is quite abstract, however at this point I'm neither able to provide an MWE nor am I allowed to share the ASN.1 definitions we use. Still I hope that this allows to triage this issue further and I'll report back with more details when I can.

brchiu:
I found there is some unsatisfied side-effect of my modification of asn1c_misc.c, so I withdraw this pull request.

PR #275, opened 2018-04-25
Modify asn_time2GT_frac to accept 9 digits of fractional seconds

rkubik:
Hopefully others also have the need for nanosecond precision in GeneralizedTime.


vlm:
Accepted, thank you!

Issue #293, opened 2018-09-10
Incorrest OER encoding of default BIT STRING value


DanyaFilatov:
Asn1c produces incorrect OER encoding when the value of the bit string field is equal to it default value.
According to the specification, when value of the field is equal to it default value, this field shall be treated as omitted.

For this type:
```
S1 ::= SEQUENCE  {
    fn INTEGER,
    fb BIT STRING DEFAULT '00'H
}
```
the S1 { 5, '00'H } is encoded as 80 01 05 00, but shall be 80 00 05

The issue is very urgent and important because out of this the COER encoding of IEEE 1609.2 certificate is invalid, and asn1c cannot be used to generate these certificates.

The actual reason is because the asn1c doesn't generate code for default values for BIT STRING sequence members:

generated code S1.c:
```
static asn_TYPE_member_t asn_MBR_S1_1[] = {
	{ ATF_NOFLAGS, 0, offsetof(struct S1, fn),
		(ASN_TAG_CLASS_CONTEXT | (0 << 2)),
		-1,	/* IMPLICIT tag at current level */
		&asn_DEF_NativeInteger,
		0,
		{ 0, 0, 0 },
		0, 0, /* No default value */
		"fn"
		},
	{ ATF_POINTER, 1, offsetof(struct S1, fb),
		(ASN_TAG_CLASS_CONTEXT | (1 << 2)),
		-1,	/* IMPLICIT tag at current level */
		&asn_DEF_BIT_STRING,
		0,
		{ 0, 0, 0 },
		0, 0, /* No default value */
		"fb"
		},
};
```
The 'fb' member definition doesn't have functions to compare and set default values. Providing these function manually, I was managed to solve the issue.
The same behavior was found for OCTET STRING.

mouse07410:
@DanyaFilatov any feedback?

@velichkov what is your opinion - is the proposed fix safe to integrate into the master branch? If so, should @brchiu create a similar patch for `OCTET STRING`?

DanyaFilatov:
For my purposes it works but I didn't have enough time to fully test other lengths, etc.. Anyway it would be great to merge it because potentially partial solution is better than no solution.
Would be also great to implement OCTET STRINGS as well.

brchiu:
With da56723a3279a3cd1d011a79a1119197030d33de, the following example asn.1 module
```
TEST

DEFINITIONS AUTOMATIC TAGS ::=

BEGIN

S1 ::= SEQUENCE  {
[...]

PR #297, opened 2018-09-27
Fix issue 293 by handling default value with type ATV_BITVECTOR

brchiu:
1. Modify asn1p_y.c and asn1p_y.y to fix an error that
   function _convert_bitstring2binary() always return NULL.

2. Add handling for default value with type ATV_BITVECTOR in
   function try_inline_default().

ref comments in #293


libasn1compiler/asn1c_C.c
```diff
@@ -2421,9 +2436,10 @@ try_inline_default(arg_t *arg, asn1p_expr_t *expr, int out) {
 		//expr->marker.flags &= ~EM_INDIRECT;
 		return 0;
 	default:
-	  if(etype & ASN_STRING_KM_MASK) {
-		if(expr->marker.default_value == NULL
-		|| expr->marker.default_value->type != ATV_STRING)
+	  if(expr->marker.default_value) {
+	    if(!(etype & ASN_STRING_KM_MASK) &&
```
mouse07410: Are you sure it should be `&&` and not `||`? The logic seems to need "if the type is not a STRING derivative, *or* default not provided, *or* (default is neither `BIT STRING` nor `OCTET STRING`), then break". Correct?

brchiu:
```
BIT STRING -> etype : ASN_BASIC_BIT_STRING , default_value_type : ATV_BITVECTOR
OCTET STRING -> etype : ASN_BASIC_OCTET_STRING, default_value_type : ATV_BITVECTOR
IA5String -> etype : ASN_STRING_IA5String, default_value_type : ATV_STRING
```
However, ASN_BASIC_BIT_STRING and ASN_BASIC_OCTET_STRING do not have bits set with ASN_STRING_KM_MASK.

I admit that this is a little obscure but I tried to be safe for possible combinations I could figure out.
To simplify it, perhaps we could just check `default_value->type` against ATV_STRING and ATV_BITVECTOR.

mouse07410: I see. Would you say that `ATV_STRING` and `ATV_BITSTRING` cover all the possibilities we want to include here?

Issue #303, opened 2018-11-19
Error encoding timestamp (INTEGER_T)


momor10:
i have a type 
TimeStampIts that's defined like this:
`TimeStampIts::= INTEGER (0..4398046511103)`
the generated type:
```
/* TimeStampIts */
typedef INTEGER_t	 TimeStampIts_t;
```

In a DENM Message im trying to set the value, encoding keeps failing:
```

    INTEGER_t timeResult;
    memset(&timeResult, 0, sizeof(timeResult));
    asn_umax2INTEGER(&timeResult, 4398046511103);
    denm_p->denm.management.detectionTime = timeResult;
...

        asn_encode_to_new_buffer_result_s result = asn_encode_to_new_buffer(0, ATS_UNALIGNED_BASIC_PER, &asn_DEF_Denm, denm_p);
        if (result.result.encoded == -1) {
            qDebug() << "Error Encoding: " << result.result.failed_type->name;
        } else {
            ...
        }
```
Does anyone have a clue on how to find out whats going wrong? Is setting the value like that the wrong way?

velichkov:
Hi @momor10,

> Does anyone have a clue on how to find out whats going wrong? 

Recompile with `-DASN_ASN_EMIT_DEBUG=1` in `CFLAGS` and provide the full `stderr` output.

Does it succeed when you set a smaller time stamp?

>     denm_p->denm.management.detectionTime = timeResult;

Copying structs that contain pointers like this seems suspicious. Try to set the value directly without using a local variable
```c
    asn_umax2INTEGER(&(denm_p->denm.management.detectionTime), 4398046511103);
```

momor10:
Thanks for the suggestion, the problem however remains.
`    asn_umax2INTEGER(&(denm_p->denm.management.detectionTime), 4398046511103);`
here the stderr output:
```
Encoding Denm as SEQUENCE (UPER)
first_extension = -1, elements = 2
About to encode ItsPduHeader
Encoding Denm->header:ItsPduHeader
Encoding ItsPduHeader as SEQUENCE (UPER)
first_extension = -1, elements = 3
About to encode ProtocolVersion
Encoding ItsPduHeader->protocolVersion:ProtocolVersion
Encoding NativeInteger ProtocolVersion 1 (UPER)
Value 1 (01/1) lb 0 ub 255 fix
Encoding integer 1 (1) with range 8 bits
[PER put 8 bits 1 to 095FCAAC+0 bits]
[...]

PR #260, opened 2018-02-20
Fix encoding unsigned 32-bit integers on 32-bit platforms

kuepe-sl:
This patch fixes a bug that was introduced with commit cc6a76b2d659c34327b15d47088faeba5579695f that prevents the encoding of 32-bit unsigned integers on platforms where `long` is 32 bits wide.

Fixes #252

PR #315, opened 2018-12-19
Fixes for StationID and TimestampITS on 32bit targets

michelm:
**Changes**
- Contains fix from pull request #260 and solves issue #303. 
- Added little test program (tests/test-32bit-targets)

**Verification**
Tested on x86_64 using CFLAGS=-m32; i.e. mimics behavior on i386. Also cross-compiled and tested on 32bit **arm** platform.

**Remarks**
Only tested for _UPER_


mouse07410:
@michelm would you like to add APER support, and create a similar PR against https://github.com/mouse07410/asn1c.git (branch `vlm_master`)?

@velichkov what is your take on this PR?

PR #327, opened 2019-03-31
Fix compiler redefinition warnings for __EXTENSIONS__

dfranusic:
* check if __EXTENSIONS__ if already
  defined


vlm:
Thank you!

PR #330, opened 2019-04-19
Respect --docdir option to ./configure

dpward:
Install documentation in the path provided when calling `./configure`. By default this is `$(datarootdir)/doc/asn1c`, which is the same path that was hard-coded in the Makefile, but packages for a distribution might follow a different convention (such as appending the version).

The files in the top-level directory that describe how to build and install the software are not typically included when installing it using `make install`.


vlm:
@dpward thank you for your contribution! Would you mind signing the Contributor License Agreement?

PR #340, opened 2019-06-06
Update asn1c-usage.tex

fei4xu:
The encode function call will crash if it's not initialized. Update the document to make it more clear.


vlm:
Thank you!

PR #346, opened 2019-08-23
Fix OCTET_STRING/BIT_STRING skeleton dependency

jonnystorm:
Missing BIT_STRING dependency in `skeletons/file-dependencies` results in uncopied Makefile dependency when OER is enabled:

```
$ cc -I. -DPDU=PDU *.c
/usr/bin/ld: /tmp/ccopqGWu.o:(.data.rel+0x38): undefined reference to `BIT_STRING_decode_oer'
/usr/bin/ld: /tmp/ccopqGWu.o:(.data.rel+0x40): undefined reference to `BIT_STRING_encode_oer'
collect2: error: ld returned 1 exit status
```

Correcting this dependency causes `BIT_STRING_oer.c` to be copied and added to `Makefile.am.libasncodec`, eliminating the above errors.


vlm:
Thank you very much!

PR #382, opened 2020-05-21
Improve test coverage for parser and AST testing.

joluxer:
I suggest to improve test coverage by including the stderr output of asn1c in the test evaluation of the test group tests-asn1c-compiler. 

After some testing and playing with the compiler itself and going through the tests, I got the impression, that this would help to improve the test quality. 

Please, take regeneration of the the reference files not too easy and do more cross checking, as I'm not sure of the current completeness of the given tests.

### DIFF
--- a/libasn1compiler/asn1c_save.c
+++ b/libasn1compiler/asn1c_save.c
@@ -234,7 +234,7 @@ asn1c__save_example_mk_makefile(arg_t *arg, const asn1c_dep_chainset *deps,
 
     for(int i = 0; i < argc; i++)
         safe_fprintf(mkf, "%s%s", i ? " " : "", argv[i]);
-	safe_fprintf(mkf, "\n\n");
+    safe_fprintf(mkf, "\n\n");
 
     fclose(mkf);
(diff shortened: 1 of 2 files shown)


libasn1compiler/asn1c_save.c
```diff
@@ -234,7 +234,7 @@ asn1c__save_example_mk_makefile(arg_t *arg, const asn1c_dep_chainset *deps,
 
     for(int i = 0; i < argc; i++)
         safe_fprintf(mkf, "%s%s", i ? " " : "", argv[i]);
-	safe_fprintf(mkf, "\n\n");
+    safe_fprintf(mkf, "\n\n");
```
vlm, line 237: Would you be open to removing this change? It has no relation to improving the test coverage.

mouse07410: While this particular one-liner doesn't affect the coverage - it still makes sense.

I recommend keeping it in, rather than dedicating a whole new PR just to fix this original identification break.

PR #389, opened 2020-07-10
Inital commit of Protobuf output

SeanCondon:
This is an example of a protobuf output from `asn1c`


mouse07410:
@SeanCondon there are some issues in your PR, such as

Screen Shot 2020-07-10 at 18 14 31

PR #419, opened 2021-03-04
Fixing support for disabled PER/OER in a couple of skeletons

gareins:
Very self explenatory


eisaev:
@gareins, I've done a similar but more complex work here: https://github.com/mouse07410/asn1c/pull/67

JFYI: mouse07410's fork is currently the most developed, and the official repository, unfortunately, is almost frozen.

PR #476, opened 2023-05-04
Disable ASN__STACK_OVERFLOW_CHECK if building with Asan enabled

pespin:


vlm:
Thank you! Fixed using a different approach (d3b5b273).

PR #512, opened 2026-02-12
asn1c: Don't emit the source filename into the generated file

hiagofranco:
This fixes several warnings à la:
	File /usr/src/debug/aktualizr-torizon/1.0+git/src/libaktualizr-posix/asn1/generated/asn1/AKDirectorMetaJson.c
in package aktualizr-torizon-src contains reference to TMPDIR (in combination with https://github.com/torizon/meta-toradex-torizon).


ukleinek:
It is unclear to me if it's Hiago or me who has to sign the CLA. The patch was created by me but Hiago committed and created the PR.

PR #515, opened 2026-05-28
Fix UPER decode issue for extensible bit string

chenjie163:
1. Fix UPER decode issue for extensible bit string.
2. Add OER encoder for open type.
3. Fix redefinition warning for asn_debug_indent.


shakespark:
Hi @chenjie163, thanks for working on this — we ran into what looks like the very same interoperability problem (extensible BIT STRING under UPER, across schema versions), so it's great to see more attention on it. A couple of observations from our debugging, in case they are useful:

**On the `BIT_STRING.c` change:** unless I'm missing a code path, I believe the added `&& !inext` guard may not change behavior. A few lines above, when the extension bit is set, `csiz` is already replaced:

```c
if(inext) {
    csiz = &asn_DEF_BIT_STRING_constraint_size;  /* effective_bits = -1 */
}
```

so by the time `if(csiz->effective_bits == 0 && !inext)` is evaluated, `inext == 1` implies `csiz->effective_bits == -1`, and the condition is already false without the new guard. Please do correct me if you observed a decode failure that this change fixes — I'd be very interested in the schema and byte stream.

When we chased the same symptom (extended BIT STRING values corrupting decode across schema versions), the root cause turned out to be on the **compiler side** rather than in the runtime decoder: for `BIT STRING (SIZE(m,...,n))` the PER range computation folds the named extension sizes into the root range, so the generated constraint table itself is wrong (`effective_bits` and bounds), and the encoder then emits a root-range encoding for values that should take the extension path — which is also where the cross-version corruption of subsequent fields comes from. No runtime guard can compensate for that, unfortunately. A fix is proposed in #541 (together with a related trailing-bit compaction issue), with encodings verified byte-for-byte against a major commercial ASN.1 toolchain — review and testing against your traffic would be very welcome.

**On the OER open type encoder:** this part looks genuinely useful — open types currently have no OER encode path at all, and wiring in `OPEN_TYPE_encode_oer` fills a real gap. Two small suggestions if you split it out (which might also make the PR easier to review, now that the `asn_debug_indent` item has landed separately via #520):

1. On the success path, `er` is returned with only `.encoded` set; `structure_ptr` and `failed_type` stay uninitialized. Using the `ASN__ENCODED_OK(er)` convention (as the other encoders do) would take care of that.
2. The commented-out `//	0, 0, ...` line in the `asn_OP_OPEN_TYPE` table could probably just be dropped.

Thanks again for pushing on this — happy to help test any revision against our cross-version regression matrix.

PR #516, opened 2026-06-03
Make check fails

Schramp:
A few changes to make  "make check" work again on this

PR #519, opened 2026-06-10
Fix ASan stack-check false positive and 32-bit PER test failure

vlm:
## Summary
- Disable the stack-depth overflow guard (`ASN__DEFAULT_STACK_MAX`) when building under AddressSanitizer. ASan inflates stack frames with redzones, so the distance-based recursion check fires false positives at normal call depth; ASan itself already catches real stack overflows.
- Guard the `check-127` PER unsigned-constraint test cases for 32-bit platforms, where the generated bounds overflow `signed long` (`LONG_MAX <= 2147483647L`), causing spurious test failures.

Both fixes are cherry-picked from #516, originally contributed by Ruud Schramp (@Schramp). Credited in `ChangeLog` and `AUTHORS`.

## Test plan
- [x] `make check` passes under ASan-instrumented build (`--enable-test-fuzzer`)
- [x] `check-127` passes on 32-bit builds (skips the overflowing constraint cases)

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -22,6 +22,7 @@ mouse07410 @ github
 Oskar Holmlund (ikarso @ github)
 Ray Zhang (zhanglei002 @ github)
 Russ Kubik <russkubik@gmail.com>
+Ruud Schramp (Schramp @ github)
 Simo Sorce <simo@redhat.com>
 theirix @ github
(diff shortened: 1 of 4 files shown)


Schramp:
Thanks for credits.

PR #522, opened 2026-06-17
Omit PER constraint references when -fno-constraints is used

Ioozzc:
With -fno-constraints, member PER/OER constraint variables are not generated, but asn_MBR_* still referenced them and caused build failures. This omits those references when constraints are disabled.


vlm:
The constraints and the constraint generation code are semantically required for PER and OER encodings. While they are not necessary (and the constraint code generation is not required) for BER and XER, it is unavoidable for PER and OER.

If `-fno-constraints` is specified, the code will produce incorrect (non-interoperable) PER and OER encoders and decoders, which would corresponds to no standard.

I have to respectfully decline the PR.

vlm:
Also see the https://github.com/vlm/asn1c/pull/527.

PR #523, opened 2026-06-17
Fix PER constraint inheritance for type aliases

Ioozzc:
Type aliases (e.g. AckCombInfo ::= CombInfo-1) had PER constraint tables generated but not linked in asn_TYPE_descriptor_t, so uPER encode/decode failed. Use the terminal type when wiring PER/OER constraints in emit_type_DEF().


vlm:
Added a test and fixed in https://github.com/vlm/asn1c/pull/526, thank you!

PR #528, opened 2026-06-22
Cut version 0.9.29

vlm:
* Fix the current status as the 0.9.29 version.


theirix:
Thank you very much for this release! It will greatly simplify Debian packaging - I'll provide a new version shortly.

PR #541, opened 2026-07-04
Fix UPER BIT STRING extension SIZE constraint folding and compaction

shakespark:
## Problem

For an extensible `SIZE` constraint with **named extension additions** — e.g. `BIT STRING (SIZE(2,...,3))` — the PER range computation (`libasn1fix/asn1fix_crange.c`) merged the extension additions into the **root** range. The emitted constraint became root `[2,3]` (effective_bits 1) instead of `[2,2]` (effective_bits 0):

```c
SIZE(2,...)   -> { APC_CONSTRAINED|APC_EXTENSIBLE, 0,0, 2,2 }   // correct
SIZE(2,...,3) -> { APC_CONSTRAINED|APC_EXTENSIBLE, 1,1, 2,3 }   // wrong (before)
SIZE(2,...,3) -> { APC_CONSTRAINED|APC_EXTENSIBLE, 0,0, 2,2 }   // correct (after)
```

So a value whose size fell in the named-extension range was encoded as if it were inside the root (extension bit 0 + a constrained length) instead of as a real extension. This was **not interoperable** with a major commercial ASN.1 toolchain (which encodes such a value identically regardless of which extension sizes a version names) and corrupted **cross-version** decoding: a decoder built from the base version (`SIZE(2,...)`, root `[2,2]`) misread the value *and* left the bit cursor in the wrong place, also corrupting the fields that followed.

On top of that, `BIT_STRING_encode_uper()` called `BIT_STRING__compactify()` unconditionally, stripping trailing 0 bits before computing `size_in_bits`. Per X.680 §22.7, trailing 0 bits are insignificant **only** for BIT STRING types with a NamedBitList; for an ordinary BIT STRING they are part of the abstract value. Under a variable/extensible SIZE constraint this silently truncated values: for `T ::= BIT STRING (SIZE(4,...,8))`, the 8-bit `'11110000'B` re-encoded as the 4-bit `'1111'B` — decode→re-encode was not idempotent. A second, related path corrupted values the same way: a value **below the constraint root's lower bound** was zero-padded up to the bound instead of encoding through the extension addition.

## Root cause

1. `libasn1fix/asn1fix_crange.c`: the union/CSV range builder merged post-`...` extension-addition elements into the PER-visible root range instead of keeping them out of it.
2. `skeletons/BIT_STRING.c` / `skeletons/OCTET_STRING.h`: `BIT_STRING_encode_uper()` compacted trailing zero bits unconditionally; the runtime had no per-type flag distinguishing a NamedBitList BIT STRING (where compaction is correct) from an ordinary one (where it corrupts the value). The same function zero-padded a below-root-lower-bound value instead of routing it through the extension addition.

## Fix

**Part 1 — constraint range.** In the union/CSV range builder, once the `...` extension marker has been seen, subsequent elements are extension additions and are not merged into the PER-visible root range.

A follow-up commit narrowed this to only the PER table emission path: the original change was unconditional inside `asn1constraint_compute_constraint_range()`, which is also shared by the generated `asn_check_constraints()` runtime checker and `asn1c_type_fits_long()`. That caused the generated checker to reject a schema's own legal extension values (`SIZE(2,...,3)` size 3, `INTEGER (0..3,...,4..7)` values 4..7, extensible `FROM()` alphabet characters) even though `uper_encode()` correctly encoded them — a checker regression. The follow-up introduces `CPR_ignore_extension_additions`, set only at the three PER-visible-root emission call sites (`emit_member_PER_constraints`) and in the `-print-constraints` diagnostic; every other consumer (the runtime checker, native C integer-width selection) keeps the pre-existing "fold additions into root" behavior. It also regenerates the `50-constraint-OK` expected outputs, which the original change had altered without updating.

**Part 2 — compaction and below-root padding**, built on top of part 1 (without it, extension-addition sizes encode inside the root and none of the reference encodings below are reproducible). `has_named_bits` is threaded from the compiler into `asn_OCTET_STRING_specifics_t` as its **last** field, so existing static initializers zero-fill it and third-party generated code keeps the (now-correct) no-compaction default:

- `skeletons/OCTET_STRING.h`: `has_named_bits` added.
- `skeletons/BIT_STRING.c`: compactify only when `specs->has_named_bits`; a below-lower-bound value now encodes through the extension addition when the constraint is extensible — matching the reference toolchain — and fails cleanly otherwise. NamedBitList types keep both compaction and padding (their trailing zeros are insignificant; the reference toolchain behaves the same, verified).
- `libasn1compiler/asn1c_C.c`: emit a per-type specifics with `has_named_bits=1` for BIT STRING types with a NamedBitList (previously all top-level BIT STRINGs shared `asn_SPC_BIT_STRING_specs`); route through `ETD_HAS_SPECIFICS`; fix the extern-declaration path for referenced types to declare `asn_OCTET_STRING_specifics_t`.
- OER `BIT_STRING_encode_oer()` does not compactify — no change needed.

## Validation

Differential testing against a major commercial ASN.1 toolchain (authoritative gold reference; asn1tools does not implement BIT STRING size extension, so it is not a cross-check here):

**Constraint range**: `SIZE(2,...,N)` now emits root `[2,2]`; encodings of every size (2..11) are byte-identical to the reference toolchain and identical across `SIZE(2,...)`, `SIZE(2,...,3)`, `SIZE(2,...,4)`; cross-version decode is correct (a value encoded by a newer version decodes back to the same bit string under the base version); a nested `SEQUENCE { a, bs BIT STRING(SIZE(2,...)), b }` recovers `a`, `bs`, and `b` where previously `bs` and the following field `b` were corrupted (self-contained regression: fixture 176, see Tests).

**Compaction / below-root**, reference toolchain = asn1c (fixed) on every vector, master shown for contrast:

| Type / value | reference = asn1c (fixed) | master |
|---|---|---|
| `T '11110000'B` | `84 78 00` | collapses to `'1111'B` |
| `T '11111110'B` | `84 7f 00` | trailing 0 dropped |
| `T '10000000'B` | `84 40 00` | dropped to 1 bit |
| `T '11110001'B` / `'11111111'B` | `84 78 80` / `84 7f 80` | OK (no trailing zeros — why the all-ones interop matrices never tripped this) |
| `T '1111'B` (root) | `78` | OK |
| `T '110'B` / `'10'B` (below root) | `81 e0` / `81 40` (extension escape) | `60` / zero-padded, value corrupted |
| `U ::= BIT STRING '11110000'B` | `08 f0` | dropped |
| `Flags {a,b,c} '110'B` and `'11000000'B` | both `2c` | unchanged — reference toolchain also strips trailing zeros for NamedBitList |

`make check` (tests-asn1c-compiler) shows no new failures vs pristine master. The fix in `asn1fix_crange.c` is type-agnostic, so the same PER-visible root computation is corrected for every constraint kind routed through that builder — INTEGER value ranges, SEQUENCE OF sizes, known-multiplier string sizes, and extensible permitted alphabets, not just BIT STRING/OCTET STRING SIZE — and that corrected constraint table is locked by fixture 175 (`175-per-range-extension-OK.asn1` + `.-EFprint-constraints`). That fixture is a compiler/constraint-table check only, the same kind of verification as fixture 167; it is not a claim that this PR re-verifies full runtime encode/decode interop (golden bytes, cross-version round trip) for INTEGER/SEQUENCE OF/string extension-region values the way it does for BIT STRING below — that runtime-level check is scoped to BIT STRING/OCTET STRING in this PR.

## Tests

- `tests/tests-asn1c-compiler/167-per-size-extension-OK.asn1` + `.-EFprint-constraints` — pins the PER-visible root `(SIZE(2,...))` for BIT STRING and OCTET STRING at the compiler level.
- `tests/tests-c-compiler/check-src/check-167.-gen-PER.c` — asserts the emitted PER table (`lb=ub=2`, `APC_EXTENSIBLE`), reference-toolchain golden encodings (`'11'B -> 0x60`, `'101'B -> 0x81 0xd0`) both directions, **and** that `asn_check_constraints()` accepts the size-3 extension value while rejecting size 4 (the checker-regression guard).
- Fixture 171 (`171-bitstring-nobits-OK.asn1` + `check-171.-gen-PER.c`): asserts the full reference-toolchain golden table plus per-vector round-trip idempotency; red on the unfixed encoder, green with the fix (ASAN/UBSAN clean). Three expected-output fixtures containing NamedBitList BIT STRINGs regenerated per repo convention (`125-bitstring-constraint-OK.asn1.-P`, `89-bit-string-enum-OK.asn1.-Pfcompound-names`/`-Pfwide-types`).
- Fixture 176 (`176-bitstring-extension-crossversion-OK.asn1` + `check-176.-gen-PER.c`) backs two independent claims in Part 2 with two separate self-contained regressions in one file (a prior version of this fixture covered only the first; the second was added to close a causal-attribution gap found during PR review — see below):
  - `check_w_roundtrip()` — the below-root-lower-bound half: encodes/decodes `{ a 5, bs '1'B (below the 2-bit root), b 7 }` through the actual generated `W` type, asserts the wire bytes match a reference-toolchain golden vector (`0b 01 87`), and asserts `a`/`bs`/`b` round-trip correctly. `bs` has no NamedBitList and its constraint's extension-addition list is empty, so this case exercises only the below-root-lower-bound zero-padding fix, not the compactify-gating fix. Verified red on this commit's direct parent (`df81efe3`) — asn1c's own encoder there emits non-interoperable `0a 83 80`, silently zero-padding `bs` into the root — and green on the fix. (An earlier version of this fixture's comment cited the older commit `8a274c3f` as the pre-fix baseline; that commit reproduces the same red bytes but predates an unrelated earlier fix in this series, so it is not the precise baseline for *this* change — corrected to cite the actual direct parent.)
  - `check_comp_roundtrip()` (new) — the compactify-gating half, isolated from the above: encodes/decodes `Comp ::= BIT STRING (SIZE(4,...,8))` (no NamedBitList) with value `'11110000'B` — the `T '11110000'B` row of the Validation table above — through a real generated type, asserting the wire bytes match the reference-toolchain golden `84 78 00` and that decode recovers the full 8-bit value. `'11110000'B` compacts to 4 bits, still inside the `[4,8]` root, so no extension routing is involved either pre- or post-fix; the wire difference is attributable to compactify alone. Verified red on `df81efe3` (asn1c's own pre-fix encoder emits `78`, which independently decodes back as the corrupted 4-bit `'1111'B`, confirmed against the reference toolchain) and green on the fix. This closes the gap where the Validation table's compaction claims (rows 1–7) had only ad hoc differential-testing backing, not a fixture.
- Fixture 175 (`175-per-range-extension-OK.asn1` + `.-EFprint-constraints`): compiler/constraint-table check that the same crange fix applies to INTEGER, SEQUENCE OF, and character-string extensible constraints (see the crange-scope note under Validation above).

Regression: `tests-c-compiler` keeps the same pre-existing FAIL/XFAIL set as pristine master with check-167 and check-171 passing on top; `tests-skeletons` keeps its known environmental FAILs unchanged; `check-parsing.sh` keeps its pre-existing diffs (unrelated duplicate-name issue in one fixture already failing on master).

Combined-integration verification (all three UPER extension fixes in this series merged on one ref): full `make check` zero new failures vs pristine master; end-to-end replay of the reporter's corpus fully green — including the key cell where a root-only 2-bit value `0x60` decodes as exactly `'11'B` under the extended v1/v2 schemas (the root is no longer widened).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


shakespark:
Follow-up on the constraint-folding fix (`asn1fix_crange.c`): since the fix lives in the type-agnostic PER constraint range builder, it affects every constraint kind with named additions after the extension marker — not just BIT STRING SIZE.

To make sure the change is safe across that whole surface, we validated the other affected kinds the same way as the original fix: OCTET STRING sizes, INTEGER value ranges (e.g. `(0..5,...,6..10)`), SEQUENCE OF sizes, and known-multiplier string sizes. For each one, encodings of extension-region values are byte-identical across schema versions and byte-identical to an independent commercial ASN.1 implementation, and cross-version decoding (old root-only schema reading new-version data, and vice versa) recovers all following fields correctly.

The last commit (cb50c85f) locks the computed constraints for those kinds as test 175, companion to test 167: the PER-visible root must not be widened by extension additions, and extensible permitted alphabets stay PER-invisible per X.691 #30.4 NOTE 2.

